PR #3, opened 2021-07-27
Fix for always showing 0 value

iStorry:
Fix for always showing 0 value
- maxParticipants
- emptyTimeout
- creationTime
- turnPassword
- enabledCodecs

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,10 +95,10 @@ svc.listRooms().then((rooms: Room[]) => {
 
 // create a new room
 const opts = {
-  name: 'myroom';
+  name: 'myroom',
   // timeout in seconds
-  emptyTimeout: 10 * 60;
-  maxParticipants: 20;
+  emptyTimeout: 10 * 60,
+  maxParticipants: 20
 }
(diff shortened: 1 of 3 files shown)


davidzhao:
I see the problem, our service based on Twirp is returning `max_participants`, but the parser is using `maxParticipants`

This file is automatically generated, so it would be overridden the next time we update them. I think a better fix would be to perform the translation [in here](https://github.com/livekit/server-sdk-js/blob/main/src/TwirpRPC.ts#L36)

iStorry:
> I see the problem, our service based on Twirp is returning `max_participants`, but the parser is using `maxParticipants`
> 
> This file is automatically generated, so it would be overridden the next time we update them. I think a better fix would be to perform the translation [in here](https://github.com/livekit/server-sdk-js/blob/main/src/TwirpRPC.ts#L36)

I understand. I found a solution to this thought had to import a whole new library for this. Please check once.

davidzhao (approved):
Thanks for the fix, this looks good!

PR #5, opened 2021-08-03
add recording options

frostbyte73:


davidzhao (approved):
looks good. I think this should work. let's test it with node-ts before merging

PR #7, opened 2021-08-07
cleaner params

frostbyte73:


davidzhao (approved):
LG, did you test with ts-node to confirm it works?

frostbyte73:
Just tested it out, it works

PR #20, opened 2022-01-04
Update README.md

dsa:
Adding a couple lines with a bit more detail on the use of express.raw middleware

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,10 @@ The JS SDK also provides helper functions to decode and verify webhook callbacks
 import { WebhookReceiver } from 'livekit-server-sdk';
 
 const receiver = new WebhookReceiver('apikey', 'apisecret');
-// assuming using express.raw
+
+// when using express.raw middleware, it's important to parse the 'application/json' content type:
+// app.use( express.raw({ type: 'application/json' }) );
+
 app.post('/webhook-endpoint', (req, res) => {
   // event is a WebhookEvent object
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -118,7 +118,10 @@ The JS SDK also provides helper functions to decode and verify webhook callbacks
 import { WebhookReceiver } from 'livekit-server-sdk';
 
 const receiver = new WebhookReceiver('apikey', 'apisecret');
-// assuming using express.raw
+
+// when using express.raw middleware, it's important to parse the 'application/json' content type:
+// app.use( express.raw({ type: 'application/json' }) );
```
davidzhao, line 123:
```suggestion
// router.use('/webhook/path', express.raw());
```

dsa: Does it make sense for this code sample to be "complete"? Meaning, adding the 2-3 lines of minimum imports (adding express and creating barebones app instance) required such that it could be copy-pasted and run? That's always my preference for code samples, but understand sometimes an abridged version is required or one that's library agnostic.

davidzhao: yeah I think we should do that. I think a good combo would be, to create a full working example in a file, and pointing to it from the README. Trying to keep things concise in the README as it needs to cover a bunch of different things

dsa: Cool, makes sense. Thanks for reviewing!

davidzhao (approved):

PR #30, opened 2022-04-06
Upgrade typedoc to fix marked dependency issue

lukasIO:
see https://github.com/livekit/server-sdk-js/security/dependabot/5


davidzhao (approved):
lg!

PR #34, opened 2022-05-07
fix: make baseUrl optional and update its commented default value

dipeshhkc:


davidzhao (approved):
looks good!

davidzhao:
actually, I misspoke. these are auto-generated protobufs.. it's not possible to pass in nil right now.

PR #42, opened 2022-06-28
Add support for segmented file output in the Egress API

biglittlebigben:


biglittlebigben:
Not sure what command generated the yarn.lock diff. Should we commit it as well of drop that change?

davidzhao (approved):

davidzhao:
> Not sure what command generated the yarn.lock diff. Should we commit it as well of drop that change?

I'd say skip on the yarn.lock change. it might be a `yarn update` or something that got ran.

PR #49, opened 2022-10-06
Update protocol

frostbyte73:
Adds Room egress and auto egress

### DIFF
--- a/src/EgressClient.ts
+++ b/src/EgressClient.ts
@@ -25,45 +25,45 @@ export interface RoomCompositeOptions {
   /**
    * egress layout. optional
    */
-  layout?: string,
+  layout?: string;
   /**
    * encoding options or preset. optional
    */
-  encodingOptions?: EncodingOptionsPreset | EncodingOptions,
+  encodingOptions?: EncodingOptionsPreset | EncodingOptions;
   /**
    * record audio only. optional
    */
-  audioOnly?: boolean,
+  audioOnly?: boolean;
   /**
    * record video only. optional
    */
-  videoOnly?: boolean,
+  videoOnly?: boolean;
   /**
    * custom template url. optional
    */
-  customBaseUrl?: string,
+  customBaseUrl?: string;
   /**
    * disable upload of json manifest file. optional
    */
-  disableManifest?: boolean,
+  disableManifest?: boolean;
 }
 
 export interface TrackCompositeOptions {
   /**
    * encoding options or preset. optional
    */
-  encodingOptions?: EncodingOptionsPreset | EncodingOptions,
+  encodingOptions?: EncodingOptionsPreset | EncodingOptions;
   /**
    * disable upload of json manifest file. optional
    */
-  disableManifest?: boolean,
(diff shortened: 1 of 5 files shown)


src/EgressClient.ts
```diff
@@ -50,6 +50,7 @@ export default class EgressClient {
    * @param audioOnly record audio only
    * @param videoOnly record video only
    * @param customBaseUrl custom template url
+   * @param disableManifest disable upload of json manifest file
    */
   async startRoomCompositeEgress(
```
davidzhao, line 55: I'm starting to think we should move these options into an object, instead of more arguments. (it's getting a bit harder to keep track of).

maybe something like
```
RoomCompositeOptions {
  encoding: EncodingOptions
  audioOnly?: boolean
  ...
}
```

We could define multiple interfaces like [here](https://github.com/livekit/client-sdk-js/blob/main/src/room/track/Track.ts#L143), for backwards compatibility.

src/EgressClient.ts
```diff
@@ -21,6 +21,51 @@ import { livekitPackage, Rpc, TwirpRpc } from './TwirpRPC';
 
 const svc = 'Egress';
 
+export interface RoomCompositeOptions {
```
davidzhao, line 24: double check to ensure these are exported from `index.ts`

src/EgressClient.ts
```diff
@@ -59,11 +105,35 @@ export default class EgressClient {
     audioOnly?: boolean,
     videoOnly?: boolean,
     customBaseUrl?: string,
+    disableManifest?: boolean,
```
davidzhao, line 108: I would remove `disableManifest` from legacy interface, and also mark it as `@deprecated`

src/EgressClient.ts
```diff
@@ -59,11 +105,35 @@ export default class EgressClient {
...
+    audioOnly?: boolean,
+    videoOnly?: boolean,
+    customBaseUrl?: string,
+    disableManifest?: boolean,
   ): Promise<EgressInfo> {
+    let layout: string;
+    if (typeof optsOrLayout === 'string') {
+      layout = optsOrLayout;
+    } else {
+      const opts = <RoomCompositeOptions>optsOrLayout;
```
davidzhao, line 124: undefined will probably crash this path

src/EgressClient.ts
```diff
@@ -91,25 +162,47 @@ export default class EgressClient {
...
+    output: EncodedFileOutput | SegmentedFileOutput | StreamOutput,
+    opts?: TrackCompositeOptions,
+  ): Promise<EgressInfo>;
   async startTrackCompositeEgress(
     roomName: string,
     output: EncodedFileOutput | SegmentedFileOutput | StreamOutput,
     audioTrackId?: string,
     videoTrackId?: string,
     options?: EncodingOptionsPreset | EncodingOptions,
+    disableManifest?: boolean,
```
davidzhao, line 178: similarly to the above, we can start moving away from the old syntax

davidzhao (approved):

PR #51, opened 2022-10-24
Add Ingress API support

biglittlebigben:


src/IngressClient.ts
```diff
@@ -0,0 +1,144 @@
...
+      participantName,
+      audio,
+      video,
+    });
+
+    const data = await this.rpc.request(
+      svc,
+      'CreateIngress',
+      req,
+      this.authHeader(participantIdentity, { roomJoin: true, room: roomName }),
```
davidzhao, line 132: should we use a different permission bit for ingress requests? since Ingress isn't tied to a room, it seems that we should require a diff perm bit.

src/IngressClient.ts
```diff
@@ -0,0 +1,144 @@
...
+    );
+    return IngressInfo.fromJSON(data);
+  }
+
+  private authHeader(participantIdentity: string, grant: VideoGrant): any {
```
davidzhao, line 137: since we seem to be copying this, I would suggest moving it to a helper function where it can be re-used.

src/authUtils.ts
```diff
@@ -0,0 +1,32 @@
+import { AccessToken } from './AccessToken';
+import { VideoGrant } from './grants';
+
+/**
+ * Utilities to handle authentication
+ */
+export class authUtils {
```
davidzhao, line 7: Sorry, I misunderstood your question earlier. If this is a class, then it'd be better to make this a super class encapsulating it. class names are always `CamelCase`. What do you think about `class ServiceBase`

where other Services simply extend off of it?

```suggestion
export class authUtils {
```

biglittlebigben: As a mostly Go developer, I do have opinions about subclassing 😊. But I'll trust the folks who actually write js.

davidzhao (approved):
nice!

PR #59, opened 2023-01-31
Replace jsonwebtoken with jose

moham96:
jose is updated more often and supports a wider range of platforms (i.e not tied to nodejs)


davidzhao:
Thank you for the PR. I like the direction of this but am concerned it will break end user's applications due to the change to async. Can you think of a way to do this without breaking the API?

moham96:
> Thank you for the PR. I like the direction of this but am concerned it will break end user's applications due to the change to async. Can you think of a way to do this without breaking the API?



> Thank you for the PR. I like the direction of this but am concerned it will break end user's applications due to the change to async. Can you think of a way to do this without breaking the API?

Well, although I think async is the right direction in generating tokens, we could try another library [jsrsasign](https://github.com/kjur/jsrsasign) should I test it and send PR?

davidzhao:
> Well, although I think async is the right direction in generating tokens, we could try another library [jsrsasign](https://github.com/kjur/jsrsasign) should I test it and send PR?

I agree that async is the right direction. the other option is to keep async, and mark the package as 2.0 so users could upgrade explicitly. @lukasIO  wdyt?

lukasIO:
I agree that switching to `jose` would be a nice upgrade. 
Bumping the server SDK version up to 2.0 also sounds reasonable in that regard!

moham96:
So should I send another pr with resolved conflicts and a version bump?

Developers75:
Any updates on this? When could it be merged?

Skaldebane:
Hi! I'm trying to use this SDK in a Cloudflare Worker, but jsonwebtoken doesn't work there.

I tried this branch and it seems to fix these issues (although it's a breaking change, needing to add `await`), but it's also based on an old commit so some other stuff (like Egress) is apparently broken.

When could this be merged under v2.0?

Code6226:
> Hi! I'm trying to use this SDK in a Cloudflare Worker, but jsonwebtoken doesn't work there.
> 
> I tried this branch and it seems to fix these issues (although it's a breaking change, needing to add `await`), but it's also based on an old commit so some other stuff (like Egress) is apparently broken.
> 

Happy to hear that Cloudflare Workers will be supported as of this PR!

Skaldebane:
@Code6226 Indeed! However there's still another issue with usage of `axios`, as `XMLHttpRequest` is apparently unsupported on Cloudflare Functions.

It's only used in one file (`TwirpRPC.ts`) and replacing it with a normal `fetch` fixes the issue! I'll try opening a pull request for this change.

davidzhao:
Closing in favor of #115

PR #62, opened 2023-03-13
multi-output egress

frostbyte73:


src/EgressClient.ts
```diff
@@ -74,6 +74,15 @@ export interface TrackCompositeOptions {
   encodingOptions?: EncodingOptionsPreset | EncodingOptions;
 }
 
+/**
+ * Used to supply multiple outputs with an egress request
+ */
+export interface EncodedOutputs {
+  fileOutput?: EncodedFileOutput | undefined;
```
davidzhao, line 81: nit: would be more economical to drop the output. i.e.

```suggestion
  file?: EncodedFileOutput | undefined;
```

In the future, we could make it `file?: EncodedFileOutput | EncodedFileOutput[]` if we wanted to handle multiple destinations.

davidzhao (approved):

src/EgressClient.ts
```diff
@@ -293,10 +293,10 @@ export class EgressClient extends ServiceBase {
     let preset: EncodingOptionsPreset | undefined;
     let advanced: EncodingOptions | undefined;
 
-    if ((<EncodedOutputs>output).fileOutput !== undefined) {
-      file = (<EncodedOutputs>output).fileOutput;
-      stream = (<EncodedOutputs>output).streamOutput;
-      segments = (<EncodedOutputs>output).segmentOutput;
+    if ((<EncodedOutputs>output).file !== undefined) {
```
davidzhao, line 296: actually, noticed an issue here. this wouldn't stick if file isn't set but only stream is. you should break those out into separate if elses..

PR #64, opened 2023-03-23
Update proto to adopt new ingress API

biglittlebigben:


noahlt (commented):
Should we also change [CreateIngressOptions](https://github.com/livekit/server-sdk-js/blob/31ea952619e3a2ceba34263eb687d0cf9a6bfc14/src/IngressClient.ts#L37) to take a `video` param rather than a `videoParams` param?

noahlt (approved):
LGTM

PR #74, opened 2023-04-21
Added canPublishSources, canUpdateOwnMetadata. updated dependencies

### LINKED ISSUE #57: Bump jsonwebtoken to 9.0.0
`jsonwebtoken` 8.5.1 has a security vulnerability: https://github.com/advisories/GHSA-27h2-hvpr-p74q. Please upgrade to 9.0.0.

davidzhao:
closes #57

### DIFF
--- a/src/grants.ts
+++ b/src/grants.ts
@@ -1,3 +1,10 @@
+export enum TrackSource {
+  CAMERA = 'camera',
+  MICROPHONE = 'microphone',
+  SCREEN_SHARE = 'screen_share',
+  SCREEN_SHARE_AUDIO = 'screen_share_audio',
+}
+
 export interface VideoGrant {
   /** permission to create a room */
   roomCreate?: boolean;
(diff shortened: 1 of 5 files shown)


src/grants.ts
```diff
@@ -1,3 +1,10 @@
+export enum TrackSource {
+  CAMERA = 'camera',
+  MICROPHONE = 'microphone',
+  SCREEN_SHARE = 'screen_share',
+  SCREEN_SHARE_AUDIO = 'screen_share_audio',
```
lukasIO, line 5: how should we handle `unknown` sources? 
list them explicitly here as well?

lukasIO (approved):

Ocupe (approved):

src/grants.ts
davidzhao: I'm not sure what the best approach for unknown is.. in terms of specifying you'd only want users to publish a track without a source identification?

perhaps we can leave it off until someone thinks of a use case for it. easier to add later on than it is to remove.

PR #84, opened 2023-06-05
Add awaitStartSignal for web egress

frostbyte73:


biglittlebigben (approved):
LGTM. Do we switch to a different linter?

frostbyte73:
> LGTM. Do we switch to a different linter?

I think the console warning is fine to just ignore

PR #93, opened 2023-07-07
Allow specifying permissions without spelling out all of the fields.

davidzhao:
Currently we are forcing users to provide all of the permission fields in an updateParticipant request. This is brittle since as new permission fields are introduced, they would require users' applications to be updated.

### DIFF
--- a/src/RoomServiceClient.ts
+++ b/src/RoomServiceClient.ts
@@ -1,10 +1,5 @@
-import {
-  DataPacket_Kind,
-  ParticipantInfo,
-  ParticipantPermission,
-  Room,
-  TrackInfo,
-} from './proto/livekit_models';
+import type { DataPacket_Kind, DeepPartial, Exact, TrackInfo } from './proto/livekit_models';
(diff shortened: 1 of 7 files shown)


Ocupe (approved):

src/RoomServiceClient.ts
```diff
@@ -223,20 +218,22 @@ export class RoomServiceClient extends ServiceBase {
    * @param permission optional, new permissions to assign to participant
    * @param name optional, new name for participant
    */
-  async updateParticipant(
+  async updateParticipant<I extends Exact<DeepPartial<ParticipantPermission>, I>>(
     room: string,
     identity: string,
     metadata?: string,
-    permission?: ParticipantPermission,
+    permission?: I,
```
lukasIO, line 225: out of curiosity, what's the intention of extending the type instead of specifying it as `    permission?: DeepPartial<ParticipantPermission>,
` directly on the property?

davidzhao: yeah.. that's a great question.. lol I don't know what I was thinking.

PR #97, opened 2023-07-26
Ensure we only replace ws to http in the scheme

davidzhao:


src/TwirpRPC.ts
```diff
@@ -24,7 +24,7 @@ export class TwirpRpc {
 
   constructor(host: string, pkg: string, prefix?: string) {
     if (host.startsWith('ws')) {
-      host = host.replace('ws', 'http');
+      host = host.replace(/^ws/, 'http');
```
lukasIO, line 27: lg! 
not sure if this is actually needed, as we check for `startsWith` already before and `string.replace` only changes the first occurrence.

davidzhao: yeah you are right.

PR #112, opened 2023-10-26
fix: do not overwrite previous grants

### LINKED ISSUE #109: AccessToken.addGrant() should not replace previous grants
Consider this following use case:
```ts

const token = new AccessToken();

token.addGrant({
 canPublish: true,
 canSubscribe: true,
 roomJoin: true,
 room: data.meetingID,
 canPublishData: true,
});

if( user.isAdmin ){
 token.addGrant({  roomAdmin: true })
}

```

This code should keep the previous grants and add `roomAdmin` grant to the token, instead, it replaces previous grants and adds only `roomAdmin` grant to the token.

I am not sure if this is intentional, I was not able to join a room with just `roomAdmin` grant.

I can open a PR that fixes this issue.

https://github.com/livekit/server-sdk-js/blob/e9e24658479c72bbcd156d62f046a51c608d8731/src/AccessToken.ts#L83

shahriar-shojib:
fixes #109

### DIFF
--- a/src/AccessToken.test.ts
+++ b/src/AccessToken.test.ts
@@ -58,3 +58,18 @@ describe('verify token is valid', () => {
     expect(decoded.video?.roomCreate).toBeTruthy();
   });
 });
+
+describe('adding grants should not overwrite existing grants', () => {
+  it('should not overwrite existing grants', () => {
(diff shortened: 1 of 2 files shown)


davidzhao (approved):
lgtm!

PR #114, opened 2023-11-28
Pull the protocol submodule forward and rebuild all of the protobufs.

spennihana:
update protocol submod to include s3 content disposition changes


lukasIO:
this has been addressed as part of https://github.com/livekit/server-sdk-js/pull/118, thanks for the PR!

PR #117, opened 2023-12-04
Replace Axios with regular `fetch`

Skaldebane:
In environments like Cloudflare Workers and Pages Functions, Axios requests don't work because `XMLHttpRequest`s are not supported.

Replacing it with a regular `fetch` fixes this.


lukasIO:
Not against this change, but we have to keep in mind that `fetch` is only available in node without flags since `v18`. 
For a v2, I guess we could also bump the minimum node requirement to v18. But we should make sure that other environments that we want to target, also support fetch ootb.

davidzhao:
should we use something like [node-fetch](https://www.npmjs.com/package/node-fetch) for node env?

lukasIO:
IIRC there were some inconsistencies between native `fetch` and `node-fetch` last time I came across it. 
Ideally we wouldn't need to include it.
I did a quick check for other popular JS runtimes and they all implement `fetch` ootb, so I guess we're fine as long as we bump nodeJS requirement to v18, which makes sense as everything else has reached EOL already.

lukasIO (approved):

davidzhao:
Thanks @lukasIO, merging this.

PR #122, opened 2023-12-08
Use globally available web crypto API instead of nodeJS crypto module

### LINKED ISSUE #121: Cloudflare Worker support - avoid using 'crypto' from Node
Just used [wrangler](https://developers.cloudflare.com/workers/wrangler/install-and-update/) locally to try the Cloudflare Worker support. I know a lot has been done recently to add support for non-Node JS environments. 

Tried it at this commit from yesterday:
https://github.com/livekit/server-sdk-js/commit/1c2e183160e611333cc69e6cb6f2238c4933281a

Here's the last thing I'm running into:

```
    src/util/server-sdk-js-main/dist/WebhookReceiver.js:1:19:
      1 │ import crypto from 'crypto';                       
        ╵                    ~~~~~~~~

  The package "crypto" wasn't found on the file system but is built into node.
```

They recommend the Web Crypto API: https://developers.cloudflare.com/workers/runtime-apis/web-crypto

After just commenting that code out locally, it was able to generate a JWT while running in wrangler - so I think this is the last thing left. Thanks!

lukasIO:
closes #121  

Also adds multiple test environments for the CI test step (`node, browser, edge-runtime`)

### DIFF
--- a/.changeset/famous-nails-shave.md
+++ b/.changeset/famous-nails-shave.md
@@ -1,0 +1,5 @@
+---
+"livekit-server-sdk": minor
+---
+
+Use globally available web crypto API instead of nodeJS crypto module
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -38,11 +38,10 @@ jobs:
       - name: Test
         run: pnpm test
   test-bun:
-    # The type of runner that the job will run on
     runs-on: ubuntu-latest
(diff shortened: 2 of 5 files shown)


davidzhao (approved):
awesome to see this!

Skaldebane:
Great to see this fixed! A workaround that worked for me was adding this to my `package.json`:

```
    "browser": {
        "crypto": false
    },
```

PR #128, opened 2024-01-11
Update README to reflect v2 changes

lukasIO:


davidzhao (approved):
lgtm,

should we place a banner at the top for folks that are migrating? To explicitly note the v1 to v2 changes.

I think it will make it possible to use it with Deno and alternative JS engines than Node.

lukasIO:
that's a good idea! I added a short migration section and a banner stating that new environments are supported. 
Let me know what you think

davidzhao (approved):
LG!

PR #132, opened 2024-02-12
Ignore unknown fields in protobuf parsing

lukasIO:
closes #131

PR #134, opened 2024-02-16
Ignore unknown fields in protobuf parsing (inbound webhooks)

ws:
Followup to #132


davidzhao (approved):
thanks! looks like we've missed a place 😢

PR #137, opened 2024-02-17
Export types needed for Egress usage with 2.0

davidzhao:


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "ES2017" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
-    "module": "ES2020" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */,
+    "module": "ESNext" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */,
```
lukasIO, line 4: Are we sure we want to update to `ESNext`?

package.json
```diff
@@ -34,6 +34,7 @@
     "@livekit/changesets-changelog-github": "^0.0.4",
     "@types/node": "^20.10.1",
     "@typescript-eslint/eslint-plugin": "^4.29.1",
+    "@types/node-jose": "^1.1.13",
```
lukasIO, line 37: `jose` ships already with types ootb. Adding another types package explicitly might just lead to some inaccuracies there. What's the intention behind adding this? (looks like the types are actually for cisco's fork of the library)

lukasIO (approved):
Exporting the types/classes makes sense.
Don't think we'd want the jose types package and the tsconfig module change.

tsconfig.json
davidzhao: it wasn't letting me use await. I can revert if there are other downsides here.

package.json
davidzhao: ah I'll revert this. the types didn't show up for me in VS code.

PR #145, opened 2024-02-29
Expose protobuf TrackSource and map TrackSource claims to string

lukasIO:
marking this as a minor version bump

### DIFF
--- a/src/grants.ts
+++ b/src/grants.ts
@@ -16,10 +16,13 @@ export function trackSourceToString(source: TrackSource) {
   }
 }
 
-export function claimsToJwtPayload(grant: ClaimGrants): JWTPayload {
-  const claim: Record<string, unknown> = { ...grant };
-  if (Array.isArray(claim.canPublishSources)) {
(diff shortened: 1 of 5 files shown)


src/grants.ts
```diff
@@ -1,10 +1,27 @@
...
+  }
+}
 
-export enum TrackSource {
-  CAMERA = 'camera',
-  MICROPHONE = 'microphone',
-  SCREEN_SHARE = 'screen_share',
-  SCREEN_SHARE_AUDIO = 'screen_share_audio',
+export function claimsToJwtPayload(grant: ClaimGrants): JWTPayload {
+  const claim: Record<string, unknown> = { ...grant };
```
davidzhao, line 20: isn't grant a `ClaimGrants`? so we'd need to check under `grant.video`?

would be great to have a unit test for this helper to ensure proper behavior.

lukasIO: yeah, thanks for catching that. Added a unit test, too

davidzhao (approved):
nice!

PR #155, opened 2024-04-16
Bump minimum engine requirement to node 19

lukasIO:
As we're using the globally available web crypto implementation in nodeJS (in order to make it work also for bun, deno and edge runtimes), we need to bump the minimum engine requirement to v19 as that's the first version that has `crypto` registered globally without the additional `--experimental-global-webcrypto` flag.


lukasIO:
By default it will only log a warning when installed. If users have a `.npmrc` file that sets `engine-strict=true` than it would error out for them with < 19

PR #173, opened 2024-04-30
Add SIP service

dennwc:


davidzhao (approved):
lgtm!

src/SipClient.ts
```diff
@@ -0,0 +1,256 @@
...
+  DeleteSIPDispatchRuleRequest,
+  SIPParticipantInfo,
+  CreateSIPParticipantRequest,
+} from '@livekit/protocol';
+import ServiceBase from './ServiceBase.js';
+import { livekitPackage, Rpc, TwirpRpc } from './TwirpRPC.js';
+import {
+  SIPDispatchRuleDirect,
+  SIPDispatchRuleIndividual,
+} from '@livekit/protocol/src/gen/livekit_sip_pb';
```
lukasIO, line 21: I think we should export this directly from the protocol package's index file, as this pattern might cause errors for some picky bundlers.

lukasIO: adding `SIPDispatchRuleDirect, SIPDispatchRuleIndividual` to the import from `@livekit/protocol` should work.

dennwc: Done!

dennwc:
@lukasIO what would be the easiest way to test it? I'm not very familiar with JS ecosystem.

lukasIO:
you could just add a `my-sip-test.ts` file in the root of your local repo and import (and use) the new APIs from there. 

You can run it directly with e.g. `pnpm dlx tsx ./my-sip-test.ts`

lukasIO (approved):

PR #186, opened 2024-06-04
Move server-sdk files into packages subdirectory

lukasIO:
In order to keep the commit history of the node-repo, we'll have to temporarily allow regular merge-commits instead of squashing them.


davidzhao:
I've enabled merge commits here

.changeset/config.json
```diff
@@ -1,11 +1,15 @@
 {
-  "$schema": "https://unpkg.com/@changesets/config@2.0.0/schema.json",
-  "changelog": ["@livekit/changesets-changelog-github", { "repo": "livekit/server-sdk-js" }],
+  "$schema": "https://unpkg.com/@changesets/config@2.2.0/schema.json",
+  "changelog": [
+    "@livekit/changesets-changelog-github",
+    {
+      "repo": "livekit/node-sdks-temp"
```
theomonnom, line 6: I guess this should be changed?

theomonnom (approved):
🚀🚀🚀
let's rename the repo to livekit/node-sdks?

lukasIO:
@nbsp I also copied over most of your README proposal for the node real-time SDK, I hope that's alright with you!

nbsp (changes requested):
this line no longer applies, we removed it in the new PR (which you should also cherry-pick over)

lukasIO:
I don't see any line associated with your comment, maybe a GH bug? 
I'm assuming you mean the `dispose` call?

nbsp:
yep

nbsp:
also, weird nit, it feels odd that our two subfolders are `rtc-node` and `livekit-server-sdk`. i know they're the package names and one of them is under a namespace — but it isn't consistent to have `livekit-` on one but not the other (and the prefix is redundant as well, maybe)

davidzhao:
can we rename `rtc-node` to `livekit-rtc` ?

COPYING
nbsp, line 1: out of `/COPYING`, `/NOTICE`, and `/LICENSE`, two of these can go, and COPYING/NOTICE are the same file
i think keeping `COPYING` is the best for GitHub UX without clashing with REUSE

lukasIO:
> can we rename rtc-node to livekit-rtc ?

sure, I'll make that change and merge the PR. 

We had users (and also Russ) confused in the past as to why repos/folders and the npm package name weren't aligned, so I thought that would be a good shout here. `livekit-rtc` doesn't make a whole lot of sense to me when the package is called `@livekit/rtc-node`.

COPYING
lukasIO: can you elaborate on the advantages UX wise? 

Does it hurt to keep `LICENSE` around anyways? It seems like pretty much every bigger repo has that file?

nbsp: so `NOTICE` isn't picked up by GitHub as being a license file at all, and doesn't give it any special UI
between `LICENSE` and `COPYING` it doesn't really matter per se but i think it's redundant to have the same file once for REUSE and once for GitHub to play nice

lukasIO: got it, thanks.

nbsp:
is GH Actions set up to release version packages for rtc-node from this repo as well or not yet?

lukasIO:
yeah, it's set up to release all packages in scope

PR #199, opened 2024-06-07
examples/basic_room: add a complete example

nbsp:
updates `tsconfig.json` to es2022 to allow toplevel `await`s

doesn't currently work due to a bug i'm having trouble figuring out:
`await room.connect(...)` exits the program immediately with code 13, lines under it do not run.
occasionally (but not always) it will throw thirty of the same error when running:

```
error calling callback status: Closing
error calling callback status: Closing
error calling callback status: Closing
# (...)
```

this error comes from `nodejs.rs`:

https://github.com/livekit/node-sdks/blob/d79d30f478bd517e8422b478b1345439635f0b49/packages/livekit-rtc/src/nodejs.rs#L30-L38

i suspect this might have something to do with the previous change to `nodejs.rs` where we made `ThreadsafeFunction` close, but it's definitely not a borrow panic since this entire closure is Rust safe, so i don't know what causes this.

as it stands i'm not sure this new `rtc-node` version is usable until this bug is fixed


pnpm-lock.yaml
```diff
@@ -1,27 +1,28 @@
-lockfileVersion: '9.0'
+lockfileVersion: '6.0'
```
lukasIO, line 1: @nbsp not very helpful, I know, but could you update your pnpm version to v9 so that this doesn't get recreated for every PR ?

nbsp: oh oops, will do

nbsp:
yeah, `tsfn.unref` is a regression. cloning it and unreferencing it after we're done doesn't do it either.

cc @theomonnom should i try the other hacks we thought up before we found unref?

nbsp:
i think i figured it out, by giving the ThreadsafeFunction to the `global` object.
this way it only gets GC'd at the end of the Node program (as opposed to immediately, like it did now),
but it won't close on any running `Room` that needs it.

using `global` is generally not advised but i think this is the one place where there's no other way to do this

theomonnom:
> ah, `tsfn.unref` is a regression. cloning it and unreferencing it after we're done doesn't do it either.

This is weird, looking at the doc of napi. Unref should have been done the work.
Did you try using [Weak = True](https://github.com/napi-rs/napi-rs/blob/3bd548d31797a78719f3c297e3df8d29824bddab/crates/napi/src/js_values/function.rs#L148)?

theomonnom:
https://github.com/napi-rs/napi-rs/blob/3bd548d31797a78719f3c297e3df8d29824bddab/crates/napi/src/threadsafe_function.rs#L305

nbsp:
i can try, but i don't think that'd do anything. we call `unref()` immediately, and in fact that's the core of the problem, from my understanding.

`unref()` lets the ThreadsafeFunction get released and subsequently dropped the moment nothing's holding it, which is the end of `livekit_initialize()`. this is what causes error 13: the function is dropped, then referenced, which causes a panic while awaiting a promise.

theomonnom:
> `unref()` lets the ThreadsafeFunction get released and subsequently dropped the moment nothing's holding it, which is the end of `livekit_initialize()`. this is what causes error 13: the function is dropped, then referenced, which causes a panic while awaiting a promise.

This isn't the case, the ThreadSafeFunction isn't dropped after livekit_intialize, [the value is moved inside the callback](https://github.com/livekit/node-sdks/blob/d79d30f478bd517e8422b478b1345439635f0b49/packages/livekit-rtc/src/nodejs.rs#L32)

nbsp:
it still gets dropped somewhere between that and running any code — within a few milliseconds of calling initialize

theomonnom (approved):
it seems like an issue from napi-rs, the new API introducing Weak isn't released yet (v3.0.0-alpha-2).
let's change to this API when this is released so we don't depend on any GC behaviour.
otherwise lgtm!

theomonnom:
Mmh actually I just tested the changes locally and it the process is still not exiting after calling disconnect.

nbsp:
> Mmh actually I just tested the changes locally and the process is still not exiting after calling disconnect. Maybe we can try still try v3.0.0 and the Weak feature? EDIT: I recommend using the git url directly, I'm seeing a bunch of commits related to ThreadSafeFunction that aren't released in any alpha version yet

can't make it work :/
i could have sworn that this fix made it work? i don't know what changed to make it not work in this version.
we're bouncing between two opposite issues: function object never getting garbage collected and not dropping ←→ function getting dropped too early and then referenced after it's gone.

could you try figuring out 3.0.0-alpha.2 yourself? there's probably something i missed here

nbsp:
@theomonnom pushed the worktree i was playing around with, a few caveats:

- napi build is broken and won't `export declare` functions when it needs to, for some reason. haven't figured it out yet, relevant only if this update works, anyway.
- sometimes, depending on how you mess with it, it's gonna throw proto errors? i don't know why or how. clearing `node_modules` and rebuilding sometimes (not always) seems to fix it.

nbsp:
should be feature complete now, hooray. still crashes prematurely though :/
i'll try tinkering with it and changing configs here and there, see what works.

PR #200, opened 2024-06-11
examples/basic_room: add a complete example

nbsp:
supersedes #199
keeping that around for the napi-rs → neon commits if we ever want to look at them.
for now, re-re-exposing `livekitDispose`, asking the user to call it directly whenever they need to.


examples/basic_room/index.ts
```diff
@@ -1,6 +1,44 @@
...
+console.log('connected to room', room);
+
+// set up audio track
+const source = new AudioSource(16000, 1);
+const track = LocalAudioTrack.createAudioTrack('audio', source);
+const options = new TrackPublishOptions();
+options.source = TrackSource.SOURCE_MICROPHONE;
+
+// read file into Uint16Array
+const sample = readFileSync(join(import.meta.dirname, '../speex.wav'));
```
theomonnom, line 37: You can add a basic wav reader on the example, it doesn't have to be fancy, just to correctly read the used sample_rate and the number of channels.

theomonnom:
Otherwise lgtm!

examples/basic_room/index.ts
```diff
@@ -36,9 +36,14 @@ options.source = TrackSource.SOURCE_MICROPHONE;
...
 const buffer = new Uint16Array(sample.buffer);
+// read relevant metadata from wav file
+// this example assumes valid encoding little-endian
+const channels = sample.readUInt16LE(22);
+const sampleRate = sample.readUInt32LE(24);
+const dataSize = sample.readUInt32LE(30);
 
 await room.localParticipant.publishTrack(track, options);
-await source.captureFrame(new AudioFrame(buffer, 16000, 1, buffer.byteLength / 2))
+await source.captureFrame(new AudioFrame(buffer, sampleRate, channels, dataSize))
```
theomonnom, line 46: the header is still sent as a frame

theomonnom: Or maybe buffer follows the cursor?

examples/basic_room/index.ts
```diff
@@ -1,6 +1,49 @@
...
+const jwt = await token.toJwt();
+
+// set up room
 const room = new Room();
-room.connect('ws://localhost:7880', 'my-token');
+await room.connect(process.env.LIVEKIT_URL, jwt, { autoSubscribe: true, dynacast: true });
+console.log('connected to room', room);
+
+// set up audio track
+const source = new AudioSource(16000, 1);
```
theomonnom, line 31: The AudioSource should also follow the right sample_rate & num_channels from the file

theomonnom (approved):
lgtm, can you add a changeset before merging

PR #204, opened 2024-06-13
Explicitly add native modules as optional dependencies

lukasIO:
adding the deps explicitly is needed now as previously the `napi prepublish` command was adding them on the fly only for the published package on npm. 
In order to work with the package more seamlessly locally, I think it makes sense to declare them explicitly as part of the workspace instead. 

as @nbsp pointed out, we'll have to keep that in mind when/if we ever add an additional build architecture and update the `optionalDependencies` field then

### DIFF
--- a/examples/basic_room/index.ts
+++ b/examples/basic_room/index.ts
@@ -1,4 +1,3 @@
-import { readFileSync } from 'node:fs';
 import {
   AudioFrame,
   AudioSource,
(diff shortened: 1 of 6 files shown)


examples/basic_room/index.ts
```diff
@@ -29,7 +33,7 @@ console.log('connected to room', room);
 
 // read relevant metadata from wav file
 // this example assumes valid encoding little-endian
-const sample = readFileSync(join(import.meta.dirname, '../speex.wav'));
+const sample = readFileSync(join(process.cwd(), './speex.wav'));
```
lukasIO, line 36: `import.meta.dirname` is in RC status and only available in NodeJS 20.11 and up.

PR #206, opened 2024-06-16
README updates

davidzhao:


nbsp:
REUSE needs that file in `LICENSES/ ` to lint against. other than that i guess this is fine for consistency and making GitHub work, though you can delete `COPYING` as it's the same as `NOTICE`

README.md
```diff
@@ -25,12 +25,12 @@ Use this SDK to add real-time video, audio and data features to your Node app. B
 ## Monorepo Navigation
 
 - **Packages**:
-  - [Server SDK](/packages/livekit-server-sdk/README.md)
-  - [Node real-time SDK](/packages/livekit-rtc/README.md)
+  - [Server SDK](https://github.com/livekit/node-sdks/tree/main/packages/livekit-server-sdk) - to interact with server APIs.
```
lukasIO, line 28: if we change the links to be absolute (I guess for npm?) then we should probably do the same for the examples further down

lukasIO (approved):

README.md
davidzhao: yeah npm was the reason for the change. However, I realized the main README isn't published to npm. So I'm rolling that portion back

PR #207, opened 2024-06-17
examples/{basic_room -> publish-wav}

nbsp:
- add more info about this example to READMEs
- stream chunked audio instead of entire file


examples/publish-wav/speex.wav
lukasIO, line 1: It's not a big file, but as it's binary we could use git lfs for this.

lukasIO (approved):
lgtm!

PR #219, opened 2024-06-21
Update protocol. Support new SIP Trunks and grants.

dennwc:


packages/livekit-server-sdk/src/SipClient.ts
```diff
@@ -81,6 +107,7 @@ export class SipClient extends ServiceBase {
   /**
    * @param number phone number of the trunk
    * @param opts CreateSipTrunkOptions
+   * @deprecated
```
lukasIO, line 110: can we add a note here that says what to use instead of the deprecated method?

packages/livekit-server-sdk/src/SipClient.ts
```diff
@@ -118,21 +145,141 @@ export class SipClient extends ServiceBase {
...
+      svc,
+      'CreateSIPOutboundTrunk',
+      req,
+      await this.authHeader({}, { admin: true }),
+    );
+    return SIPOutboundTrunkInfo.fromJson(data, { ignoreUnknownFields: true });
+  }
+
+  /**
+   * @deprecated
```
lukasIO, line 248: also a note for what to use instead here

.changeset/short-jokes-promise.md
```diff
@@ -0,0 +1,11 @@
+---
+'@livekit/rtc-node-linux-arm64-gnu': minor
+'@livekit/rtc-node-win32-x64-msvc': minor
+'@livekit/rtc-node-linux-x64-gnu': minor
+'@livekit/rtc-node-darwin-arm64': minor
+'@livekit/rtc-node-darwin-x64': minor
+'livekit-server-sdk': minor
+'@livekit/rtc-node': minor
```
lukasIO, line 8: I don't think this change affects the `@livekit/rtc-node` packages (yet)? 
we should remove all lines apart from `livekit-server-sdk`

lukasIO (commented):
couple of nits

.changeset/short-jokes-promise.md
dennwc: Yes, I was not sure why it marked it as changed. It's probably because of lock file. I'll remove them from the changeset then.

dennwc:
Done!

lukasIO (approved):

PR #221, opened 2024-06-21
Adds missing await to startParticipantEgress

lourd:
Using the `startParticipantEgress` method on the EgressClient always returns a 401 because the auth header's not actually being set, due to a missing `await`.

Confirmed this was the issue by trying the same request with the livekit-cli which was capable of starting a participant egress without issue.


davidzhao (approved):
great catch. this was a miss when we updated the protocol to v2

PR #222, opened 2024-06-24
Export RoomEgress for AutoEgress support

davidzhao:
We didn't previously export `RoomEgress`, so it wasn't possible to use AutoEgress with the SDK


lukasIO (approved):
lgtm

PR #239, opened 2024-07-17
Update rust sdk, add dtmf and attributes support

lukasIO:
depends on https://github.com/livekit/rust-sdks/pull/371

and pulls in latest changes on main of rust sdk, which comes with a couple of necessary ffi protocol changes.

~~the submodule currently points to the `lukas/add-publish-dtmf` branch. Should update that after that PR got merged to point to main instead. ~~ ✅


lukasIO:
@nbsp also added support for retrieving attribute events (which I think wasn't yet part of #240 ?)

packages/livekit-rtc/src/utils.ts
```diff
@@ -0,0 +1,25 @@
+// SPDX-FileCopyrightText: 2024 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+export function diffAttributes(
```
theomonnom, line 5: Ah awesome!
cc @keepingitneil, we should do the same for https://github.com/livekit/python-sdks/pull/217

theomonnom (approved):
lgtm! 🔥

PR #242, opened 2024-07-22
Add ability to update attributes

lukasIO:
supersedes #241


davidzhao (approved):
nice, this is better!

PR #255, opened 2024-08-17
Update protocol to support SRT output

davidzhao:


packages/livekit-server-sdk/src/index.ts
```diff
@@ -16,6 +16,8 @@ export {
   EncodingOptions,
   EncodingOptionsPreset,
   GCPUpload,
+  ImageCodec,
+  ImageFileSuffix,
```
lukasIO, line 20: just double checking that these two exports are the only thing we need for enabling SRT output?

davidzhao: these are unrelated to SRT. StreamProtocol should have SRT now though

packages/livekit-server-sdk/src/EgressClient.ts
```diff
@@ -1,28 +1,31 @@
 // SPDX-FileCopyrightText: 2024 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-import {
+import type {
```
davidzhao, line 4: @lukasIO I'm not sure if this is a good idea to change.. but my editor did it automatically.. should I revert?

nbsp: this is fine, there's no functional change, prettier enforces this for clarity

davidzhao: ok, give it a ✅ if it looks fine. we'll get it released

nbsp (approved):

PR #265, opened 2024-09-11
publishTranscription

bcherry:


.changeset/nasty-ravens-live.md
```diff
@@ -0,0 +1,5 @@
+---
+"@livekit/rtc-node": patch
```
nbsp, line 2:
```suggestion
"@livekit/rtc-node": minor
```

? cc @theomonnom your call

packages/livekit-rtc/src/participant.ts
```diff
@@ -130,6 +135,34 @@ export class LocalParticipant extends Participant {
     }
   }
 
+  async publishTranscription(transcription: Transcription): Promise<void> {
```
nbsp, line 138: nit: no need for `: Promise<void>`

packages/livekit-rtc/src/transcription.ts
```diff
@@ -0,0 +1,14 @@
+export interface TranscriptionSegment {
+  id: string;
+  text: string;
+  startTime: number;
+  endTime: number;
```
nbsp, line 5: should these be BigInts here to begin with?

nbsp (commented):
\+ reuse header on transcription.ts

packages/livekit-rtc/src/transcription.ts
bcherry: yeah that makes sense. of course they're all 0 everywhere anyways!

.changeset/nasty-ravens-live.md
bcherry: yeah you're probably right. i changed that but am happy to change back

nbsp (approved):
looks good, just run `format:write` to fix tests

PR #271, opened 2024-09-18
rtc: bump ffi to v0.10.2

nbsp:


theomonnom (approved):
Nice, can you add the new parameters inside the AudioSource. (buffer_size_ms)
There is also:
wait_for_playout
clear_queue

nbsp:
@theomonnom can you re-check? want to make sure i did this impl correctly

packages/livekit-rtc/src/audio_source.ts
```diff
@@ -47,7 +63,46 @@ export class AudioSource {
     this.ffiHandle = new FfiHandle(res.source.handle.id);
   }
 
+  get queuedDuration(): number {
+    return Math.max(this.currentQueueSize - Date.now() + this.lastCapture, 0);
```
theomonnom, line 67: Date.now isn't monotonic, can you use https://nodejs.org/api/process.html#process_process_hrtime_bigint

packages/livekit-rtc/src/utils.ts
```diff
@@ -0,0 +1,32 @@
+// SPDX-FileCopyrightText: 2024 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import EventEmitter, { once } from 'events';
+
+/** @internal */
+export class Queue<T> {
```
theomonnom, line 7: Seems like the Queue isn't needed, just Promise/Future?

nbsp: this promise is reused more than one time so it's more ergonomic to use queue for this

packages/livekit-rtc/src/audio_source.ts
```diff
@@ -47,7 +63,49 @@ export class AudioSource {
...
   async captureFrame(frame: AudioFrame) {
+    const now = Number(process.hrtime.bigint() / 1000000n);
+    const elapsed = this.lastCapture === 0 ? 0 : now - this.lastCapture;
+    this.currentQueueSize += (frame.samplesPerChannel / frame.sampleRate - elapsed) * 1000;
+
+    // remove 50ms to account for processing time (e.g. using wait_for_playout for very small chunks)
+    this.currentQueueSize -= 50;
+    this.lastCapture = now;
+
+    if (this.timeout) {
```
theomonnom, line 104: Seems like this.timeout is never set?

Is `waitForPlayout` working correctly?

nbsp: oops yeah. it seemed to work fine enough so i didn't notice.

packages/livekit-rtc/src/audio_source.ts
```diff
@@ -47,7 +63,50 @@ export class AudioSource {
...
+    });
+
+    FfiClient.instance.request<ClearAudioBufferResponse>({
+      message: {
+        case: 'clearAudioBuffer',
+        value: req,
+      },
+    });
+
+    this.releaseQueue.put();
```
theomonnom, line 85: if clearQueue is called multiple times, it seems like waitForPlayout will directly returns?

theomonnom (changes requested):

packages/livekit-rtc/src/audio_source.ts
nbsp: right this isn't one-to-one. i'll rework this to use Futures

davidzhao (approved):

theomonnom (approved):

PR #272, opened 2024-09-19
{Audio,Video}Stream: use AsyncIterableIterator

nbsp:
instead of having a typed EventEmitter

### DIFF
--- a/packages/livekit-rtc/src/audio_stream.ts
+++ b/packages/livekit-rtc/src/audio_stream.ts
@@ -7,6 +7,7 @@ import { FfiClient, FfiClientEvent, FfiHandle } from './ffi_client.js';
 import type { AudioStreamInfo, NewAudioStreamResponse } from './proto/audio_frame_pb.js';
 import { AudioStreamType, NewAudioStreamRequest } from './proto/audio_frame_pb.js';
 import type { Track } from './track.js';
+import { Mutex } from './utils.js';
 
 export class AudioStream implements AsyncIterableIterator<AudioFrame> {
   /** @internal */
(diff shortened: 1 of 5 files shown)


packages/livekit-rtc/src/audio_stream.ts
```diff
@@ -70,15 +59,36 @@ export class AudioStream extends (EventEmitter as new () => TypedEmitter<AudioSt
...
+  async next(): Promise<IteratorResult<AudioFrame>> {
+    if (this.eventQueue.length > 0) {
+      const value = this.eventQueue.shift();
+      if (value) {
+        return { done: false, value };
+      } else {
+        return { done: true, value: undefined };
+      }
+    }
+    return new Promise((resolve) => (this.queueResolve = resolve));
```
lukasIO, line 83: hm. I think `this.queueResolve` might get overridden before being called?
so the async iterator could get stuck

lukasIO: the `Mutex` helper that Théo has mentioned already might make sense here

nbsp: could it? i don't think that's the case, it only gets overridden if you call `next()` twice without awaiting the first one, which you shouldn't do anyway.

lukasIO: > which you shouldn't do anyway.

agreed that it _shouldn't_ be done, but seems like a footgun regardless?

nbsp: would it make sense to return a copy of the same promise if `next` is called twice?

lukasIO: I'm not sure, I think best case scenario is a promise chain, which is exactly what the `Mutex` class does.

theomonnom (approved):

theomonnom:
Can you do the same for VideoStream?

PR #276, opened 2024-09-20
RPC implementation via FFI

bcherry:
This is a full implementation of the [native RPC spec ](https://www.notion.so/livekit/RPC-Proposal-435567131ce64077aa4466fafc2fb407?pvs=4) based on Rust FFI https://github.com/livekit/rust-sdks/pull/461

It includes documentation and a complete working example.


bcherry:
Thinking more about this - the ack being user facing is an unfortunate side effect of using event emitters because you need to know not just that the SDK received the call but also the actual app was configured to listen. 

This could be fixed if we forced you to register for RPC explicitly, that way the framework could send ack automatically. And you'd get something a bit more serious-feeling than event listeners. I'll make that change next week

examples/rpc/index.ts
```diff
@@ -0,0 +1,174 @@
...
+
+const performGreeting = async (room: Room): Promise<void> => {
+  console.log('[Requester] Letting the greeter know that I\'ve arrived');
+  try {
+    const response = await room.localParticipant!.performRpcRequest('greeter', 'arrival', 'Hello');
+    console.log(`[Requester] That's nice, the greeter said: "${response}"`);
+  } catch (error) {
+    console.error('[Requester] RPC call failed:', error);
+    throw error;
+  }
```
bcherry, line 94: pretty simple calling interface

examples/rpc/index.ts
```diff
@@ -0,0 +1,174 @@
...
+}
+
+const registerReceiverMethods = async (greetersRoom: Room, mathGeniusRoom: Room): Promise<void> => {
+  greetersRoom.localParticipant?.registerRpcMethod(
+    'arrival',
+    async (request: RpcRequest, sender: RemoteParticipant) => {
+      console.log(`[Greeter] Oh ${sender.identity} arrived and said "${request.payload}"`);
+      await new Promise((resolve) => setTimeout(resolve, 2000));
+      return "Welcome and have a wonderful day!";
+    },
```
bcherry, line 64: seems like a dead simple way to declare support for an RPC method? you can throw in here too

packages/livekit-rtc/src/participant.ts
```diff
@@ -266,6 +279,169 @@ export class LocalParticipant extends Participant {
...
+    const maxRoundTripLatencyMs = 2000;
+
+    return new Promise((resolve, reject) => {
+      const id = crypto.randomUUID();
+
+      const request = new RpcRequest({
+        id,
+        method,
+        payload,
+        responseTimeoutMs: responseTimeoutMs - maxRoundTripLatencyMs
```
bcherry, line 312: this seems like a decent enough approximation

packages/livekit-rtc/src/participant.ts
```diff
@@ -266,6 +279,169 @@ export class LocalParticipant extends Participant {
...
+      });
+
+      const responseTimeoutId = setTimeout(() => {
+        this.pendingResponses.delete(id);
+        reject(new Error(RPC_ERROR_RESPONSE_TIMEOUT));
+      }, responseTimeoutMs);
+
+      this.pendingResponses.set(id, (response) => {
+        if (this.pendingAcks.has(id)) {
+          console.error('RPC response received before ack', id);
```
bcherry, line 342: not really much of a problem

packages/livekit-rtc/src/rpc.ts
```diff
@@ -0,0 +1,58 @@
+// TODO: This implementation is only a prototype
+//       The final version will use a protocol types where possible
+export class RpcRequest {
+  id: string;
+  method: string;
+  payload: string;
```
bcherry, line 6: in the future I'd like to be able to extend payload to support streaming large payloads - haven't thought enough about whether this interface would be backwards compatible when that happens or not

boks1971: I am thinking a separate type would be cleaner. Something like `RpcRequestStream`.

boks1971 (approved):
Looks great @bcherry

packages/livekit-rtc/src/rpc.ts
bcherry: makes sense - do we think it would be ok for it to be a whole parallel API (e.g. `registerStreamingRpcMethod`, `performStreamingRpcMethod`)?

boks1971: That would be my take. The separation is very clear that way.

lukasIO (approved):
nice!

bcherry:
Updated with more robust error handling and unit tests. Still need to go native after merging protocol PR but I think this is now done, otherwise. @lukasIO mind taking a look over the error handling to see how you like it?

packages/livekit-rtc/src/rpc.ts
```diff
@@ -0,0 +1,165 @@
...
+    UNCAUGHT_ERROR: 1001,
+    UNSUPPORTED_METHOD: 1002,
+    CONNECTION_TIMEOUT: 1003,
+    RESPONSE_TIMEOUT: 1004,
+    RECIPIENT_DISCONNECTED: 1005,
+    PAYLOAD_TOO_LARGE: 1006,
+    MALFORMED_RESPONSE: 1007,
+  } as const;
+
+  static ErrorMessages = {
```
lukasIO, line 109: feels like error codes and error message should be coupled under a common key?

packages/livekit-rtc/src/rpc.ts
```diff
@@ -0,0 +1,165 @@
...
+ */
+export function byteLength(str: string): number {
+  const encoder = new TextEncoder();
+  return encoder.encode(str).length;
+}
+
+/**
+ * @internal
+ */
+export function truncateBytes(str: string, maxBytes: number): string {
```
lukasIO, line 150: maybe log a warning if a string got truncated?

packages/livekit-rtc/src/rpc.ts
```diff
@@ -0,0 +1,165 @@
...
+ * serialized and sent across the wire. The sender will receive an equivalent error on the other side.
+ *
+ * Build-in types are included but developers may use any string, with a max length of 256 bytes.
+ */
+
+export class RpcError extends Error {
+  static MAX_MESSAGE_BYTES = 256;
+  static MAX_DATA_BYTES = 15360; // 15 KB
+
+  code: number;
```
lukasIO, line 72:
```suggestion
  code: ErrorCodes | number;
```

lukasIO (approved):

packages/livekit-rtc/src/rpc.ts
bcherry: yeah - how do we handle logging in this SDK? we don't really do it right now. ok to use console.warn here?

packages/livekit-rtc/src/rpc.ts
bcherry: doesn't seem to work like that - i'm a bit confused by these module-enums tbh. leaving as a number, I think its fine

packages/livekit-rtc/src/rpc.ts
bcherry: i added `@internal` to the messages - the idea is to expose the codes in the SDK but not the messages, so devs can do `if error.code == RpcError.ErrorCode.UNSUPPORTED_METHOD` etc...

packages/livekit-rtc/src/rpc.ts
lukasIO: a dedicated and configurable logger would make sense, I think, e.g. https://github.com/pinojs/pino

packages/livekit-rtc/src/rpc.ts
lukasIO: Yeah that makes sense, but we could define the `messages` with e.g. `Record<keyof ErrorCode, string>` so that they're typed with the keys of `ErrorCode` being the only source of truth

packages/livekit-rtc/src/rpc.ts
lukasIO: ah, probably it was `RpcError.ErrorCodes | number`

packages/livekit-rtc/src/rpc.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  toJSON() {
+    return {
+      code: this.code,
+      message: this.message,
+      data: this.data,
+    };
+  }
+
+  static ErrorCode = {
```
lukasIO, line 97: can we device some sort of number based grouping for those already?
would e.g. be nice to have RECIPIENT_DISCONNECTED and RECIPIENT_NOT_FOUND in the same group.

bcherry: yeah sounds good, I'll group them thematically into 10xx, 11xx, 12xx etc in the final version.

packages/livekit-rtc/src/rpc.ts
bcherry: yeah it didn't like that for some reason. I'll circle back on this in the final version, thank you!

packages/livekit-rtc/src/rpc.ts
bcherry: yeah i might leave logs out of the final version

packages/livekit-rtc/src/rpc.ts
bcherry: 🧙‍♂️

packages/livekit-rtc/src/rpc.ts
bcherry: done

packages/livekit-rtc/src/rpc.ts
bcherry: I went with 14xx and 15xx so they have some vague association to http

packages/livekit-rtc/README.md
```diff
@@ -76,10 +76,57 @@ await room.localParticipant.publishTrack(track, options);
 await source.captureFrame(new AudioFrame(buffer, 16000, 1, buffer.byteLength / 2));
 ```
 
+### RPC
+
+RPC you to perform your own predefined method calls from one participant to another. This feature is especially powerful when used with [Agents](https://docs.livekit.io/agents), for instance to forward LLM function calls to your client application.
```
nbsp, line 81: nit: typo

packages/livekit-rtc/README.md
```diff
@@ -76,10 +76,57 @@ await room.localParticipant.publishTrack(track, options);
...
+
+LiveKit is a dynamic realtime environment and calls can fail for various reasons. 
+
+You may throw errors of the type `RpcError` with a string `message` in an RPC method handler and they will be received on the caller's side with the message intact. Other errors will not be transmitted and will instead arrive to the caller as `1500` ("Application Error"). Other built-in errors are detailed in `RpcError`.
+
 ## Examples
 
-- [`publish-wav`](https://github.com/livekit/node-sdks/tree/main/examples/publish-wav): connect to a room and publish a wave file
+- [`publish-wav`](https://github.com/livekit/node-sdks/tree/main/examples/publish-wav): connect to a room and publish a .wave file
+- [ `rpc`](https://github.com/livekit/node-sdks/tree/main/examples/rpc): simple back-and-forth RPC interaction
```
nbsp, line 127: nit: space

packages/livekit-rtc/src/participant.ts
```diff
@@ -266,6 +293,151 @@ export class LocalParticipant extends Participant {
...
+   * @param payload - The method payload
+   * @param responseTimeoutMs - Timeout for receiving a response after initial connection
+   * @returns A promise that resolves with the response payload or rejects with an error.
+   * @throws Error on failure. Details in `message`.
+   */
+  async performRpcRequest(
+    destinationIdentity: string,
+    method: string,
+    payload: string,
+    responseTimeoutMs?: number,
```
nbsp, line 310: drop "ms" here and everywhere else, node assumes times and durations are in milliseconds by default

nbsp (approved):

packages/livekit-rtc/src/participant.ts
bcherry: I'm going to leave this as-is since this feature directly crosses language boundaries I'd like to keep the interface as consistent as possible with the wire protocol. this field will be called `response_timeout_ms` in every SDK.

lukasIO (approved):

packages/livekit-rtc/src/participant.ts
```diff
@@ -349,6 +371,163 @@ export class LocalParticipant extends Participant {
...
+   *
+   * @param method - The name of the RPC method to unregister
+   */
+  unregisterRpcMethod(method: string) {
+    this.rpcHandlers.delete(method);
+
+    const req = {
+      localParticipantHandle: this.ffi_handle.handle,
+      method,
+    } as UnregisterRpcMethodRequest;
```
lukasIO, line 477: same as above

packages/livekit-rtc/src/participant.ts
```diff
@@ -349,6 +371,163 @@ export class LocalParticipant extends Participant {
...
+      payload: string,
+      responseTimeoutMs: number,
+    ) => Promise<string>,
+  ) {
+    this.rpcHandlers.set(method, handler);
+
+    const req = {
+      localParticipantHandle: this.ffi_handle.handle,
+      method,
+    } as RegisterRpcMethodRequest;
```
lukasIO, line 459: would be good to avoid this type cast and use the protobuf creation scheme instead. 

```
req: RegisterRpcMethodRequest = create(RegisterRpcMethodRequestSchema, { ... });

File: packages/livekit-rtc/src/participant.ts [hunk 15]
```diff
@@ -349,6 +371,163 @@ export class LocalParticipant extends Participant {
...
+        }
+      }
+    }
+
+    const req = {
+      localParticipantHandle: this.ffi_handle.handle,
+      invocationId,
+      error: responseError ? responseError.toProto() : undefined,
+      payload: responsePayload ?? undefined,
+    } as RpcMethodInvocationResponseRequest;
```
lukasIO, line 521: same

packages/livekit-rtc/src/rpc.ts
```diff
@@ -0,0 +1,84 @@
...
+  static fromProto(proto: RpcError_Proto) {
+    return new RpcError(proto.code, proto.message, proto.data);
+  }
+
+  toProto() {
+    return {
+      code: this.code as number,
+      message: this.message,
+      data: this.data,
+    } as RpcError_Proto;
```
lukasIO, line 42: same

packages/livekit-rtc/src/participant.ts
bcherry: yep - thank you

packages/livekit-rtc/src/participant.ts
bcherry: yep - thank you

packages/livekit-rtc/src/rpc.ts
bcherry: yep - thank you

PR #278, opened 2024-09-25
Add chat API support

lukasIO:


packages/livekit-rtc/src/participant.ts
```diff
@@ -185,6 +190,60 @@ export class LocalParticipant extends Participant {
     });
   }
 
+  async sendChatMessage(
+    text: string,
+    destinationIdentities?: Array<string>,
```
bcherry, line 195: does this default to all? Worth adding docstrings

packages/livekit-rtc/src/participant.ts
```diff
@@ -185,6 +190,60 @@ export class LocalParticipant extends Participant {
...
+    if (cb.error) {
+      throw new Error(cb.error);
+    }
+    return cb.chatMessage!;
+  }
+
+  async editChatMessage(
+    editText: string,
+    originalMessage: ChatMessage,
+    destinationIdentities?: Array<string>,
```
bcherry, line 222: seems kinda weird that you can give different values here than to the original. this might also cause participants who join a call after an initial message is sent but before an edit to receive the edit and maybe have strange behavior? not sure if there's an easy way to solve though

bcherry (approved):

packages/livekit-rtc/src/participant.ts
lukasIO: yeah, had the same feeling and opted to omit destinationIdentites on client-sdk-js altogether for now because of that

packages/livekit-rtc/src/types.ts
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2024 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+export interface ChatMessage {
+  id: string;
+  timestamp: number;
+  message: string;
+  editTimestamp?: number;
+  generated?: boolean;
+}
```
nbsp, line 10: as far as i understand this is the same as the proto interface but with bigints swapped out for numbers? there's already a bunch of conversions happening in code and i wonder if this cuts down on enough of them to warrant this utility type or whether it's okay to just use the proto one

nbsp (approved):
lg, seconding ben's comments plus this one

packages/livekit-rtc/src/types.ts
lukasIO: are you concerned about performance reasons of the conversions or maintenance burden?
reasons for me to keep them separate: 
- we might be adding stuff in the proto that we don't necessarily want to expose to the user facing API in the same form
- it's a bit of a pain to work with bigints in JS

lukasIO: >  just a bit weird how this is the first time we've had to do this

FWIW in the client js sdk we do this for almost every proto type

nbsp:
also add a changeset

PR #286, opened 2024-10-03
feat(rtc): add AudioResampler, combineAudioFrames

nbsp:


packages/livekit-rtc/src/audio_resampler.ts
```diff
@@ -0,0 +1,166 @@
...
+      return [];
+    }
+
+    const outputData = FfiClient.instance.copyBuffer(res.outputPtr, res.size);
+    return [
+      new AudioFrame(
+        new Int16Array(outputData.subarray()),
+        this.#outputRate,
+        this.#channels,
+        outputData.length,
```
theomonnom, line 124: This doesn't seems right, samples per channel != bytes count 
You should divide by 2 and num_channels

packages/livekit-rtc/src/audio_resampler.ts
```diff
@@ -0,0 +1,166 @@
...
```
theomonnom, line 162: Same here

theomonnom (changes requested):

packages/livekit-rtc/src/audio_frame.ts
```diff
@@ -52,3 +52,42 @@ export class AudioFrame {
...
+    }
+
+    if (frame.channels != channels) {
+      throw new Error(`channel mismatch: expected ${channels}, got ${frame.channels}`);
+    }
+
+    totalSamplesPerChannel += frame.samplesPerChannel;
+  }
+
+  const data = new Int16Array(buffer.map((x) => [...x.data]).flat());
```
theomonnom, line 91: Should we preallocate instead?

nbsp: i don't think node is at a level where it cares about preallocation vs conversion

packages/livekit-rtc/src/audio_resampler.ts
nbsp: looking at our own example this seems to be fine:

https://github.com/livekit/node-sdks/blob/main/examples/publish-wav/index.ts#L60

```ts
     Math.trunc(frameSize / channels),

theomonnom: I if you check above, data size was already divided by two

bcherry (commented):
this could be a good candidate for unit tests. we don't have them in the agents framework yet but it would be good to have them in the base SDK as a start. 

I added the scaffold in https://github.com/livekit/node-sdks/pull/287

theomonnom (approved):

PR #287, opened 2024-10-03
Add support for tests in livekit-rtc, fix linter warnings

bcherry:
I pulled this out of #276 so we can start adding tests to the rtc package.  To verify it works, I also wrote some tests for Mutex and even found a bug (I think) with multi locking (not that we use it).

While I was here I addressed all of the linter warnings, although some of the server stuff with explicit anys I decided to just ignore rather than try to reason about...


nbsp (approved):
while you're here could you split the "build lint and test" CI into its components for better readability? https://github.com/livekit/agents-js/pull/26/files

bcherry:
@nbsp updated - lmk if this is what you had in mind

nbsp:
yep, awesome. feel free to merge

PR #296, opened 2024-10-11
Fix ingress creation parameter inline documentation

biglittlebigben:


packages/livekit-server-sdk/src/IngressClient.ts
```diff
@@ -22,11 +22,11 @@ export interface CreateIngressOptions {
...
   /**
-   * name of the room to send media to. optional
+   * name of the room to send media to.
    */
   roomName?: string;
   /**
-   * unique identity of the participant. optional
+   * unique identity of the participant.
    */
   participantIdentity?: string;
```
davidzhao, line 31: since they are not optional, we shouldn't make them `?`. and should move these to the top?

biglittlebigben: Would that potentially break any existing code using the SDK?

davidzhao: if this is required, then existing code would either have to set it, or it wouldn't be successful either way, right?

packages/livekit-server-sdk/src/IngressClient.ts
```diff
@@ -67,11 +67,11 @@ export interface UpdateIngressOptions {
    */
   name: string;
   /**
-   * name of the room to send media to. optional
+   * name of the room to send media to.
```
davidzhao, line 70: this is optional right? on updates.

biglittlebigben: Everything is optional in the dictionary in the sense that parameters left out will not be updated. It's odd to single out these 2 parameters. We could flag all parameters as optional, but this seems redundant on a parameter called "UpdateIngressOptions".

davidzhao: got it.. makes sense.

packages/livekit-server-sdk/src/IngressClient.ts
```diff
@@ -22,11 +22,11 @@ export interface CreateIngressOptions {
...
   /**
-   * name of the room to send media to. optional
+   * name of the room to send media to. required
    */
   roomName?: string;
   /**
-   * unique identity of the participant. optional
+   * unique identity of the participant. required
    */
   participantIdentity?: string;
```
davidzhao, line 31:
```suggestion
   */
  participantIdentity: string;
```

packages/livekit-server-sdk/src/IngressClient.ts
```diff
@@ -146,17 +146,27 @@ export class IngressClient extends ServiceBase {
...
-      video = opts.video;
-      participantMetadata = opts.participantMetadata;
+    if (opts == null) {
+      throw new Error('options dictionary is required');
+    }
+
+    name = opts.name || '';
+    roomName = opts.roomName;
+    participantName = opts.participantName || '';
+    participantIdentity = opts.participantIdentity || '';
```
davidzhao, line 156:
```suggestion
    participantIdentity = opts.participantIdentity;
```

packages/livekit-server-sdk/src/IngressClient.ts
```diff
@@ -146,17 +146,27 @@ export class IngressClient extends ServiceBase {
...
-      enableTranscoding = opts.enableTranscoding;
-      url = opts.url || '';
-      audio = opts.audio;
-      video = opts.video;
-      participantMetadata = opts.participantMetadata;
+    if (opts == null) {
+      throw new Error('options dictionary is required');
+    }
+
+    name = opts.name || '';
```
davidzhao, line 153: need consts in front of these

PR #305, opened 2024-10-17
Send SDK override and version to FFI

bcherry:
Screenshot showing this coming out correctly in SFU logs:

image

This PR depends on these three PRs, all merged.  Just waiting on FFI bump, but planning to do RPC at the same time.
https://github.com/livekit/rust-sdks/pull/471
https://github.com/livekit/livekit/pull/3110
https://github.com/livekit/cloud/pull/3083


packages/livekit-rtc/src/nodejs.rs
```diff
@@ -13,10 +12,16 @@ use napi::{
 };
 use napi_derive::napi;
 use prost::Message;
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
+
+static SDK_VERSION: OnceLock<String> = OnceLock::new();
```
theomonnom, line 17: is there a reason to store it statically?

theomonnom (approved):
awesome lgtm!

packages/livekit-rtc/src/nodejs.rs
bcherry: No it seems to work just as well with const. I don't think it makes any difference but I updated it anyways.

bcherry:
@nbsp maybe you can take a look at the way i'm getting the version. Originally I had it importing package.json directly at runtime but that didn't seem to work reliably so I have it now generating a file at build time. I guess ideally the changeset script would generate this file but I read the docs and couldn't find an obvious way to do that.

nbsp:
idea: add the version file to gitignore, generate it on build/release time on the same step it publishes to npm. that way it pulls in the version file at runtime without redundancy.

see how livekit/protocol JS directory doesn't have protobufs, but the npm package does.

packages/livekit-rtc/src/nodejs.rs
bcherry: actually I can't figure out why I thought I needed to store this here anyways. it's being stored in the server's FfiConfig and that's all it needs.

bcherry:
@nbsp yeah good idea - that seems to work fine and will prevent weird local churn.

PR #303, opened 2024-10-16
Enable strict type checking

lukasIO:
This also fixes all errors that showed up and paves the way for more ergonomic usage by upgrading protobuf-es to v2.x ([once the protovalidate types are available ](https://github.com/bufbuild/protobuf-es/issues/966))

PR #308, opened 2024-10-22
fix example and comment for protobuf-es-2

bcherry:
some cleanup outside the critical path from https://github.com/livekit/node-sdks/pull/303/files

1. Fix the comment in ./generate_proto.sh to show the correct dependency version
2. Fix the wav publish example to use the new TrackPublishOptions type


examples/publish-wav/index.ts
```diff
@@ -41,9 +41,10 @@ const dataSize = sample.readUInt32LE(40) / 2;
 // set up audio track
 const source = new AudioSource(sampleRate, channels);
 const track = LocalAudioTrack.createAudioTrack('audio', source);
-const options = new TrackPublishOptions();
+const options = {
+  source: TrackSource.SOURCE_MICROPHONE,
+} as TrackPublishOptions;
 const buffer = new Int16Array(sample.buffer);
-options.source = TrackSource.SOURCE_MICROPHONE;
```
bcherry, line 46: @lukasIO I was a little surprised this was broken and wonder if it means that we aren't exporting the API compatibility helpers you added in the way you expected? I didn't dig in but it might be worth a quick look

lukasIO: The helpers are exposed, I forgot to initialise the contructor arguments with an empty object to allow for passing no params at all 🤦 , sorry! 

Fixed in #310

packages/livekit-rtc/generate_proto.sh
```diff
@@ -5,7 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # This script requires protobuf-compiler and https://www.npmjs.com/package/@bufbuild/protoc-gen-es
-# `brew install protobuf-c && npm install -g @bufbuild/protoc-gen-es@1.10.0`
+# `brew install protobuf-c && npm install -g @bufbuild/protoc-gen-es@2.2.0`
```
bcherry, line 8: @lukasIO can you just make this change in your fix PR https://github.com/livekit/node-sdks/pull/310

PR #310, opened 2024-10-23
TS strict mode fixes

lukasIO:


bcherry (approved):
FWIW I'm fine with just merging this and discarding the revert

PR #322, opened 2024-11-04
fix(proto): don't crash on plain JavaScript

nbsp:
unsure what `fileDesc` is.


lukasIO (approved):
thanks 🙏

lukasIO:
> unsure what fileDesc is.

It's protobuf-es v2 related: https://github.com/bufbuild/protobuf-es/issues/969

PR #334, opened 2024-11-13
chore: support CommonJS modules

nbsp:


packages/livekit-rtc/package.json
```diff
@@ -56,7 +65,7 @@
   },
   "scripts": {
     "prebuild": "node -p \"'export const SDK_VERSION = ' + JSON.stringify(require('./package.json').version) + ';'\" > src/version.ts",
-    "build:tsc": "pnpm prebuild && tsc && cp -r src/napi dist/",
+    "build:tsc": "pnpm prebuild && tsup --onSuccess \"tsc --declaration --emitDeclarationOnly\" && cp -r src/napi dist/",
```
lukasIO, line 68: `tsc` is misleading here, not sure what to change it to though.

nbsp: maybe just ts?

packages/livekit-rtc/tsconfig.json
```diff
@@ -1,12 +1,12 @@
 {
   "compilerOptions": {
-    "module": "ES2020",
+    "module": "ES2015",
```
lukasIO, line 3: what's the oldest node version we want to support?

tsup.config.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const defaultOptions: Options = {
+  entry: ['src/index.ts'],
+  format: ['cjs', 'esm'],
+  splitting: false,
+  sourcemap: true,
+  // for the type maps to work, we use tsc's declaration-only command
+  dts: false,
+  clean: true,
+  target: 'es6',
```
lukasIO, line 11: I think we could the target more explicit by writing `node16` here

packages/livekit-rtc/package.json
lukasIO: also: tsup docs suggest to use `tsup-node` to invoke for building node libs

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "ES2017" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
-    "module": "ES2020" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */,
+    "target": "es6" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
+    "module": "esnext" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */,
```
lukasIO, line 4: what is this tsconfig used for? 
`esnext` doesn't seem like the right option for anything 🤔

nbsp: oh i think i changed this by accident to figure out the 2015 issue. i'll revert

packages/livekit-rtc/tsconfig.json
nbsp: unsure but if it works with es2015 it'll work with all of them. before es2015 there was no const/let, so it's not worth going back any more

packages/livekit-rtc/package.json
nbsp: that didn't work for me for some reason.

packages/livekit-rtc/package.json
```diff
@@ -56,9 +65,9 @@
   },
   "scripts": {
     "prebuild": "node -p \"'export const SDK_VERSION = ' + JSON.stringify(require('./package.json').version) + ';'\" > src/version.ts",
-    "build:tsc": "pnpm prebuild && tsc && cp -r src/napi dist/",
-    "build": "pnpm build:tsc && napi build --platform --release --dts native.d.ts --js native.cjs --pipe \"prettier -w\" src/napi",
-    "artifacts": "pnpm build:tsc && napi artifacts",
+    "build:ts": "pnpm prebuild && tsup --onSuccess \"tsc --declaration --emitDeclarationOnly\" && cp -r src/napi dist/ && cp -r src/napi/* dist/",
```
nbsp, line 68: this line is silly but it's to get around esbuild's issues with dynamic import of fs

packages/livekit-rtc/package.json
```diff
@@ -56,9 +65,9 @@
...
+    "build": "pnpm build:ts && napi build --platform --release --dts native.d.ts --js native.cjs --pipe \"prettier -w\" src/napi",
```
lukasIO, line 69: this looks like napi is only building for cjs, is that the behaviour we want?

nbsp: we have a wrapper for esm, napi only generates  CJS

nbsp: update: this doesn't actually work, tried using agents and it broke. unsure where to go from here because napi doesn't do ESM

lukasIO: what's the error you're running into?

nbsp: something about not being able to find the `native.cjs` file when running it under cjs. can't reproduce right now, i'm having a different issue related to tsup on Agents in specific

nbsp: upon further testing with node-sdks examples directly this doesn't appear to be reproduceable 🤷‍♀️

packages/livekit-rtc/package.json
lukasIO: is the order of things correct here? it looks like it first copies napi files but napi build is only called afterwards in the `build` script 👀

lukasIO: hm it was like this also before, just not sure I understand the reason 🤷

nbsp: there's a few files in src/napi already pre-generation, namely the shim to translate them from CJS to ESM if needed. i do think this might be a bit more convoluted than necessary but it works and doesn't seem to take too long on this part

lukasIO (approved):

PR #336, opened 2024-11-18
chore(tsconfig): enable `noUncheckedIndexedAccess`

nbsp:
only added to livekit-rtc since server-sdk has some other issues compiling and strict is disabled on it regardless


lukasIO (approved):
would be nice to get this (and strict mode) for server-sdk-js, too!

PR #337, opened 2024-11-20
Add missing createSIPParticipant params

umarniz:
The following params exist in the Go SDK but not in the node SDK:

```
- RingingTimeout
- MaxCallDuration
- EnableKrisp
```

They were added last month to the protocol: https://github.com/livekit/protocol/commit/19b686d3128927e025eb9d5258c636cc0de3e790

Tested in my local project to work correctly.

### DIFF
--- a/packages/livekit-server-sdk/src/SipClient.ts
+++ b/packages/livekit-server-sdk/src/SipClient.ts
@@ -423,9 +423,9 @@ export class SipClient extends ServiceBase {
     let playRingtone: boolean = false;
     let playDialtone: boolean = false;
     let hidePhoneNumber: boolean = false;
-    let ringingTimeout: number = 0;
-    let maxCallDuration: number = 0;
-    let enableKrisp: boolean = false;
+    let ringingTimeout: number | undefined = undefined;
+    let maxCallDuration: number | undefined = undefined;
+    let enableKrisp: boolean | undefined = undefined;
 
     if (opts !== undefined) {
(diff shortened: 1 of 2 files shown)


umarniz:
Added a change set as I assume there will be a minor version bump with this change. Feel free to take over the PR as a maintainer or ping me if I can help with something.

umarniz:
Pinging @dennwc for review as I saw you reviewed another SIP PR.

Hope not the wrong reviewer 🙏🏽

dennwc (approved):

packages/livekit-server-sdk/src/SipClient.ts
```diff
@@ -441,6 +451,9 @@ export class SipClient extends ServiceBase {
       playRingtone: playDialtone,
       playDialtone: playDialtone,
       hidePhoneNumber: hidePhoneNumber,
+      ringingTimeout: new Duration({ seconds: BigInt(ringingTimeout) }),
+      maxCallDuration: new Duration({ seconds: BigInt(maxCallDuration) }),
```
davidzhao, line 455: when it's not passed in, we should leave these as `undefined` instead of set to 0. this way the server is able to pick sane defaults, versus having different defaults from each client

umarniz: @davidzhao thanks for the review, defaulted the 3 new params to be undefined if not specified

davidzhao (approved):
thanks for the PR!

PR #342, opened 2024-11-28
Update protocol dependency

lukasIO:


nbsp (approved):

nbsp:
maybe remove agent-dispatch example from changeset though

lukasIO:
It's not part of the changeset, but it will get tagged for a release as it has the server-sdk as a dependency.

PR #349, opened 2024-11-30
support node v18

lalitkapoor:
Extract out the digest method to support other implementations. Use this as a fallback if the web crypto api is not available. As a result we're now able to support node versions < 19. Node 18 is still officially supported via maintenance related updates, and so setting the minimum supported version to node v18.


lalitkapoor:
@lukasIO let me know if this looks good.

.github/workflows/ci.yml
```diff
@@ -70,14 +70,17 @@ jobs:
 
   test:
     name: Test
+    strategy:
+      matrix:
+        node-version: [18, 20]
```
lukasIO, line 75: That's good thinking, let's add 22 as well while we're at it.

packages/livekit-server-sdk/src/digest.ts
```diff
@@ -0,0 +1,15 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Use the Web Crypto API if available, otherwise fallback to Node.js crypto
+export default async function digest(algorithm: string, data: ArrayBuffer): Promise<ArrayBuffer> {
+  if (globalThis.crypto?.subtle) {
+    return globalThis.crypto.subtle.digest(algorithm, data);
+  } else {
+    const nodeCrypto = await import('node:crypto');
+    const hash = nodeCrypto.createHash(algorithm.replace('-', '').toLowerCase());
```
lukasIO, line 11: I don't believe we're likely to change the algorithm here in the foreseeable future. I'd rather go with hard coded algorithm names here and above instead of the string manipulation.

packages/livekit-server-sdk/src/WebhookReceiver.ts
```diff
@@ -67,7 +68,7 @@ export class WebhookReceiver {
       const claims = await this.verifier.verify(authHeader);
       // confirm sha
       const encoder = new TextEncoder();
-      const hash = await crypto.subtle.digest('SHA-256', encoder.encode(body));
+      const hash = await digest('SHA-256', encoder.encode(body));
```
lukasIO, line 71: instead of converting this back and forth, we can pass the body directly and only do the textencoding part in the global webcrypto path

.github/workflows/ci.yml
lalitkapoor: added `22` and `latest` for future proofing

lalitkapoor:
@lukasIO ready for another review!

ivanma9:
Hey! I downgraded my node to use v18.x in my package.json. I still get the `node: crypto` error
```
	"engines": {
		"node": "18.x"
	},
```

Screen Shot 2025-01-16 at 1 01 44 PM

lukasIO:
@ivanma9 where are you running this? Please post the full logs instead of a screen shot

PR #351, opened 2024-12-06
fix(deps): move @bufbuild/protobuf to dependencies

danielmahon:
Move `@bufbuild/protobuf` to dependencies as its used in runtime and can conflict with newer versions (2.x)

2.x versions no longer export the WKT classes and because we are using 2.x in the parent app `livekit-server-sdk` will resolve to using that version throwing this error:

```
main-1  | file:///usr/src/app/node_modules/livekit-server-sdk/dist/SipClient.js:1
main-1  | import { Duration } from "@bufbuild/protobuf";
main-1  |          ^^^^^^^^
main-1  | SyntaxError: The requested module '@bufbuild/protobuf' does not provide an export named 'Duration'
main-1  |     at ModuleJob._instantiate (node:internal/modules/esm/module_job:146:21)
main-1  |     at ModuleJob.run (node:internal/modules/esm/module_job:229:5)
main-1  |     at ModuleLoader.import (node:internal/modules/esm/loader:473:24)
main-1  | 
```

(for a quick fix we patched this file to use the new 2.x DurationSchema creation method)


lukasIO (approved):
great catch, thanks for this!

PR #358, opened 2024-12-11
Fix CJS package masquerading as ESM on import

### LINKED ISSUE #357: Using ESM-style imports from `livekit-server-sdk` gives warnings ("current file is a CommonJS module")
I'm trying to migrate a Redwood 8 project's livekit sdks from v1 to v2. Technically it all worked fine, but there seems to be something off with the bundling. No problem on the web side (`livekit-client`), but on *each and every* `import` line `from 'livekit-server-sdk'` VS Code is giving red squiggles:

> The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("livekit-server-sdk")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts', or add the field `"type": "module"` to '/home/phil/prog/interview-tool/interviewtool/api/package.json'.ts(1479)

I'm aware (as the error message suggest) that i can change the current project's tsconfig.json, but what bugs me is that i can import fine from all other external libraries (using exclusively `import … from` at the top of the files).

Checking on https://publint.dev/livekit-server-sdk@2.9.3 gives:

grafik

→ https://publint.dev/rules#export_types_invalid_format

Checking with https://arethetypeswrong.github.io/?p=livekit-server-sdk%402.9.3 confirms this is basically a compatibility issue with the way the ESM module is declared:

grafik

The resolution is explained here: https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseESM.md

Philzen:
May fix #357

How to generate the required `.d.cts` using tsup or tsc went over the top of my head, however `npx publint` advised that it would suffice to duplicate the `.d.ts` file for that (see commit message of b5d562a2f14ba13d2eff4fe6c7eff9872c7de804 for a detailed explanation).

### DIFF
--- a/packages/livekit-server-sdk/package.json
+++ b/packages/livekit-server-sdk/package.json
@@ -20,7 +20,7 @@
     "src"
   ],
   "scripts": {
-    "build": "tsup --onSuccess \"tsc --declaration --emitDeclarationOnly\"",
+    "build": "tsup --onSuccess \"tsc --declaration --emitDeclarationOnly && cp ./dist/index.d.ts ./dist/index.d.cts\"",
     "build:watch": "tsc --watch",
     "build-docs": "typedoc",
     "changeset": "changeset",
--- a/packages/livekit-server-sdk/package.json
+++ b/packages/livekit-server-sdk/package.json
@@ -11,7 +11,6 @@
   "type": "module",
   "exports": {
     ".": {
-      "types": "./dist/index.d.ts",
       "import": "./dist/index.js",
       "require": "./dist/index.cjs"
     }


Philzen:
I've added `livekit-server-sdk` into the project that exposed #357 as a local path using this branch … and can confirm that the warning is gone :laughing:

lukasIO:
thanks for this, I didn't see it and opened #359 which uses tsup directly to generate these, so we don't have to manually copy the files around. Thanks for raising the issue and offering a fix!

Philzen:
> opened #359 which uses tsup directly to generate these, so we don't have to manually copy the files around. 

That's of course the way (and feels much more trustworthy meaning less hacky). So at least i learned what i was looking for was hidden behind the `dts: true` option all along :wink:

PR #361, opened 2024-12-17
Add data stream receiving

lukasIO:


examples/data-streams/README.md
```diff
@@ -0,0 +1,39 @@
+# RPC Example
```
bcherry, line 1: Data Streams!

bcherry (changes requested):
Can you add docstrings to all of the exported classes? There's also at least one error handling TODO left in here still.

Looks good and thank you for including an example!

examples/data-streams/README.md
lukasIO: 🙈

lukasIO:
> Can you add docstrings to all of the exported classes? 

added some, the annoying thing however is that docstrings don't show up on class instances. And we're only returning class instances to users, so they won't really profit from them. 

> There's also at least one error handling TODO left in here still

errors are currently being logged to console, the TODO is in there for later improvements around re-requesting missing chunks or similar repair mechanisms. Deleted the TODOs

packages/livekit-rtc/src/participant.ts
```diff
@@ -207,6 +223,222 @@ export class LocalParticipant extends Participant {
...
+        const textInBytes = new TextEncoder().encode(textChunk);
+
+        if (textInBytes.byteLength > STREAM_CHUNK_SIZE) {
+          this.abort?.();
+          throw new Error('chunk size too large');
+        }
+
+        return new Promise(async (resolve) => {
+          // FIXME we need an equivalent for this on the rust layer
+          // await localP.engine.waitForBufferStatusLow(DataPacket_Kind.RELIABLE);
```
lukasIO, line 279: @theomonnom How does Rust currently handle the datachannel buffer? Do we have an event yet and/or a different mechanism for backpressure?

theomonnom: No, we don't have a similar mechanism on Rust yet, what will happen is that publish_data is going to fail (returning an error)

packages/livekit-rtc/src/data_streams/stream_writer.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  constructor(writableStream: WritableStream<T>, onClose?: () => void) {
+    this.writableStream = writableStream;
+    this.defaultWriter = writableStream.getWriter();
+    this.onClose = onClose;
+  }
+
+  write(chunk: T): Promise<void> {
+    return this.defaultWriter.write(chunk);
+  }
```
nbsp, line 21: possible syntactic sugar: array expansion? i feel like `([a, b])` is ugly and doesn't make things any clearer either way, so maybe being being able to do `a, b?` would be nicer. this would require some typescript witchcraft to make it not break file streams though, so i'm fine either way

lukasIO: yeah, agree, I had tried that, but couldn't make it work with TypeScript :/ If you have an idea how it could work, let me know!

bcherry (approved):

nbsp (approved):

.changeset/many-suns-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@livekit/rtc-node": patch
```
nbsp, line 2: should be minor

lukasIO: hm, why? 
we're still in 0.x so minor versions are more likely to introduce breaking changes, which this doesn't

nbsp: this is a new API

examples/data-streams/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  const identity = 'tester';
+  const token = await createToken(identity, roomName);
+
+  const room = new Room();
+
+  const finishedPromise = new Promise((resolve) => {
+    room.on(RoomEvent.ParticipantDisconnected, resolve);
+  });
+
+  room.on(RoomEvent.TextStreamReceived, async (reader: TextStreamReader) => {
```
bcherry, line 47: what would happen if the developer actually registered two different handlers for TextStreamReceived? would they work correctly?

examples/data-streams/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  streamWriter?.close();
+};
+
+const sendFile = async (room: Room, recipient: RemoteParticipant) => {
+  console.log('sending file');
+  await room.localParticipant?.sendFile('./assets/maybemexico.png', {
+    destinationIdentities: [recipient.identity],
+    fileName: 'mex',
+    mimeType: 'image/png',
+  });
```
bcherry, line 32: i'd add `topic` here as well, and maybe in the other spot to, just to better show how to to use it.

examples/data-streams/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  for (const c of greeting) {
+    await streamWriter?.write(c);
+  }
+
+  streamWriter?.close();
+};
+
+const sendFile = async (room: Room, recipient: RemoteParticipant) => {
+  console.log('sending file');
+  await room.localParticipant?.sendFile('./assets/maybemexico.png', {
```
bcherry, line 28: this file doesn't exist, but when i added one and ran the example i got a rust panic

```
sending file
thread '<unnamed>' panicked at src/nodejs.rs:55:13:
failed to handle request: invalid request: message is empty
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
```

examples/data-streams/index.ts
bcherry: i think we'd be better off not putting this in RoomEvent at all then, and instead exposing a single spot you can register a single handler...

packages/livekit-rtc/src/participant.ts
lukasIO: this has been implemented on rust

PR #362, opened 2024-12-20
Fix screenshare option not being applied to participant egress

chrisackermann:
The screenShare option on ParticipantEgressOptions is not being passed through to the ParticipantEgressRequest, so it's never applied when initiating the egress.  This is preventing screenshare recordings when using participant egress.


davidzhao (approved):
great catch!

davidzhao:
could you add a changeset please?

chrisackermann:
> could you add a changeset please?

added!

PR #365, opened 2024-12-27
add close method to AudioSource and VideoSource

davidzhao:
ensures we can clean up allocated resources correctly


davidzhao:
porting over @typester's fix [on Python](https://github.com/livekit/python-sdks/pull/336) to Node.

packages/livekit-rtc/src/audio_source.ts
```diff
@@ -134,4 +134,8 @@ export class AudioSource {
       throw new Error(cb.error);
     }
   }
+
+  close() {
```
lukasIO, line 138: just a thought, if we ever need to do additional cleanups here, we might want to make this async from the start so that it's not breaking once we need to wait for a rust layer callback for something

lukasIO (approved):

packages/livekit-rtc/src/audio_source.ts
typester: Making it async for future flexibility makes sense to me. Since it's a client-side library, the slight overhead introduced by making it async unnecessarily at this point should be acceptable.

davidzhao: good call for future-proof. will change it to async

PR #368, opened 2024-12-29
revert upgrade to protobuf-es 2

nbsp:
This reverts commit f204721, which introduces various bugs with version mismatches between protobuf-es 1.x and 2.x. see https://github.com/livekit/agents-js/issues/232

would appreciate as many eyes on this as possible, to:
- make sure i didn't revert any newly added changes as well
- audit ***all*** uses of `!` (i think it's safe to use when necessary, due to optional/required, but worth checking)
- comprehend the jank

still not done with room.ts because it is big and messy


packages/livekit-rtc/src/audio_source.ts
```diff
@@ -14,14 +13,14 @@ import type {
...
-  NewAudioSourceRequestSchema,
+  CaptureAudioFrameRequest,
+  ClearAudioBufferRequest,
+  NewAudioSourceRequest,
 } from './proto/audio_frame_pb.js';
 
 export class AudioSource {
   /** @internal */
-  info: AudioSourceInfo;
+  info?: AudioSourceInfo;
```
lukasIO, line 23: in what situation would this be missing?

packages/livekit-rtc/src/audio_stream.ts
```diff
@@ -1,18 +1,17 @@
...
 import { FfiClient, FfiClientEvent, FfiHandle } from './ffi_client.js';
 import type { AudioStreamInfo, NewAudioStreamResponse } from './proto/audio_frame_pb.js';
-import { AudioStreamType, NewAudioStreamRequestSchema } from './proto/audio_frame_pb.js';
+import { AudioStreamType, NewAudioStreamRequest } from './proto/audio_frame_pb.js';
 import type { Track } from './track.js';
 
 export class AudioStream implements AsyncIterableIterator<AudioFrame> {
   /** @internal */
-  info: AudioStreamInfo;
+  info?: AudioStreamInfo;
```
lukasIO, line 14: same question as above

packages/livekit-rtc/src/participant.ts
```diff
@@ -61,40 +58,40 @@ import type { ChatMessage } from './types.js';
 
 export abstract class Participant {
   /** @internal */
-  info: ParticipantInfo;
+  info?: ParticipantInfo;
```
lukasIO, line 61: same question

packages/livekit-rtc/tsconfig.json
```diff
@@ -2,7 +2,16 @@
...
+    "declarationDir": "dist",
+    "module": "ES2020",
+    "esModuleInterop": true,
+    "allowJs": true,
+    "declaration": true,
+    "declarationMap": true,
+    "target": "es2020",
+    "lib": ["es2020"],
+    "moduleResolution": "node",
+    "sourceMap": true,
```
lukasIO, line 14: why do we need all these additions on the package level?

lukasIO (approved):

packages/livekit-rtc/tsconfig.json
nbsp: this is a reverting artifact from before we added all of them at a workspace level, will remove

packages/livekit-rtc/src/audio_source.ts
nbsp: unsure, but in general i prefer to use `?` over `!` if the former works. could be swayed to replace it with an exclamation point later in the file

lukasIO: if we expect an `AudioSource` to always have a `AudioSourceInfo` (which I think we do as it's required in the FFI protocol), I think we should reflect that on the user facing API.

packages/livekit-rtc/package.json
```diff
@@ -46,17 +46,17 @@
     }
   },
   "dependencies": {
-    "@bufbuild/protobuf": "^2.2.0",
+    "@bufbuild/protobuf": "^1.4.2",
```
lukasIO, line 49:
```suggestion
    "@bufbuild/protobuf": "^1.10.0",
```

packages/livekit-rtc/package.json
```diff
@@ -46,17 +46,17 @@
...
   },
   "devDependencies": {
-    "@bufbuild/protoc-gen-es": "^2.2.0",
     "@napi-rs/cli": "^2.18.0",
     "@types/node": "^20.9.2",
     "prettier": "^3.0.3",
     "tsup": "^8.3.5",
-    "typescript": "^5.2.2"
+    "typescript": "^5.2.2",
+    "@bufbuild/protoc-gen-es": "^1.4.2"
```
lukasIO, line 59:
```suggestion
    "@bufbuild/protoc-gen-es": "^1.10.0"
```

lukasIO (approved):
lgtm!

davidzhao (commented):
🥳

PR #372, opened 2025-01-03
feat(rtc): expose DisconnectReason

nbsp:


packages/livekit-rtc/src/participant.ts
```diff
@@ -95,6 +100,10 @@ export abstract class Participant {
   get kind(): ParticipantKind | undefined {
     return this.info.kind;
   }
+
+  get disconnectReason(): DisconnectReason | undefined {
+    return this.info.disconnectReason;
```
davidzhao, line 105: info.disconnectReason might be set to unknown (since not nullable)

we'd want to check and return undefined if so

lukasIO (commented):
I think it would be nice if this was part of the `ParticipantDisconnected` event payload as a second argument

nbsp:
> I think it would be nice if this was part of the `ParticipantDisconnected` event payload as a second argument

it's already sent as part of the participant, so wouldn't it be redundant? i can see how it might be nice though

davidzhao (approved):
lgtm

PR #376, opened 2025-01-06
fix(e2ee): fix types

nbsp:


packages/livekit-rtc/src/e2ee.ts
```diff
@@ -259,7 +259,8 @@ export class E2EEManager {
     const options = { ...defaultE2EEOptions, ...opts };
 
     this.options = options;
-    this.keyProvider = new KeyProvider(roomHandle, options.keyProviderOptions);
+    this.keyProvider =
```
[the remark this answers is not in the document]
nbsp: i'm not sure it will be, but if we don't set it here, we have to either:
1. assert its existence with `!`, which has broken things already
2. not assume it exists, even though it definitely does

out of the two this is the less-likely-to-break-things option

packages/livekit-rtc/src/e2ee.ts
```diff
@@ -259,7 +259,8 @@ export class E2EEManager {
...
+      options.keyProviderOptions && new KeyProvider(roomHandle, options.keyProviderOptions);
```
lukasIO, line 263: but this line here sets keyProvider to false/undefined if there are no keyProviderOptions passed. That doesn't seem right. There should always be a key provider, even if it's initialized with the default options

nbsp: ah, i see. the default values were ignored outside of this function. latest commit runs the e2ee options through `...defaultE2EEOptions,` before continuing in RoomOptions

lukasIO (approved):

lukasIO:
changeset is still missing

Issue #378, opened 2025-01-06
Error when calling unpublishTrack


s-hamdananwar:
Calling `unpublishTrack` results in an error since `stop_on_publish` bool value is not set in the function.

PR #379, opened 2025-01-06
fix(participant): added stopOnUnpublish value to UnpublishTrackRequest

s-hamdananwar:
fixes https://github.com/livekit/node-sdks/issues/378

### DIFF
--- a/packages/livekit-rtc/src/participant.ts
+++ b/packages/livekit-rtc/src/participant.ts
@@ -369,11 +369,11 @@ export class LocalParticipant extends Participant {
     }
   }
 
-  async unpublishTrack(trackSid: string) {
+  async unpublishTrack(trackSid: string, stopOnUnpublish?: boolean) {
     const req = new UnpublishTrackRequest({
       localParticipantHandle: this.ffi_handle.handle,
       trackSid: trackSid,
-      stopOnUnpublish: true,
+      stopOnUnpublish: stopOnUnpublish ?? true,
     });
 
     const res = FfiClient.instance.request<UnpublishTrackResponse>({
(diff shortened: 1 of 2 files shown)


packages/livekit-rtc/src/participant.ts
```diff
@@ -373,6 +373,7 @@ export class LocalParticipant extends Participant {
     const req = new UnpublishTrackRequest({
       localParticipantHandle: this.ffi_handle.handle,
       trackSid: trackSid,
+      stopOnUnpublish: true,
```
lukasIO, line 376: nice, we should also add this as an optional parameter in the public method that defaults to `true`

s-hamdananwar: @lukasIO is there any time we would want it to be set as `false`, assuming the user might pass it in? we did the same in python sdk as well in [this pr](https://github.com/livekit/python-sdks/pull/325)  by hardcoding it to true without giving an option to the users. let me know what you think and i can update Python sdk as well accordingly if needed

lukasIO: the use case is for users to unpublish tracks without stopping them, which enables the re-use of tracks.

[We do have this option in the browser SDK ](https://github.com/livekit/client-sdk-js/blob/main/src/room/participant/LocalParticipant.ts#L1224-L1227) for that reason, I think it would make sense for node and python, too.

s-hamdananwar: sounds good, i will update both

lukasIO (approved):

nbsp (approved):

PR #382, opened 2025-01-07
Fix cjs usage

### LINKED ISSUE #381: (0 , import_digest.default) is not a function TypeError: (0 , import_digest.default) is not a function
It seems commonjs compatibility might still have some issues. 
I tried using WebhookReceiver - which partially works. 
When hitting the sha validation of the body i get:
`(0 , import_digest.default) is not a function TypeError: (0 , import_digest.default) is not a function`

Unfortunately, I did not find the time to inspect the cjs building stuff here throughly but maybe someone involved can spot the issue.
I had a quick look at the transpiled code and suspiciously only the digest has a `__toESM()` while other imports are untouched... 

```js
// in WebhookReceiver.cjs
// ... line 35
module.exports = __toCommonJS(WebhookReceiver_exports);
var import_protocol = require("@livekit/protocol");
var import_AccessToken = require("./AccessToken.cjs");
var import_digest = __toESM(require("./digest.cjs"), 1);
const authorizeHeader = "Authorize";
```

lukasIO:
closes #381

For some reason the default export gets wrongly marked as ESM for CJS. Converting the import to a named import changes the output in dist to 
```
var import_protocol = require("@livekit/protocol");
var import_AccessToken = require("./AccessToken.cjs");
var import_digest = require("./digest.cjs");
```

### DIFF
--- a/.changeset/healthy-pots-design.md
+++ b/.changeset/healthy-pots-design.md
@@ -1,0 +1,5 @@
+---
+"livekit-server-sdk": patch
+---
+
+Fix cjs usage of WebhookReceiver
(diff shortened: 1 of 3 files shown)


nbsp:
potentially worth auditing/removing all default exports because they don't play well with esm/cjs

PR #410, opened 2025-02-05
Expose sipNumber when creating an Sip Participant

BramRoets:
This allows you to define the sipNumber (=fromNumber) when placing an outbound call.

This is required when dealing with outbound trunks where multiple numbers are configured.


davidzhao (approved):
lgtm, thanks for fixing these.

could you merge from main and add a changeset with `pnpm changeset`. I'll merge it right after

BramRoets:
Fixed in https://github.com/livekit/node-sdks/pull/427/files

PR #414, opened 2025-02-13
update to krispEnabled for consistency

davidzhao:


packages/livekit-server-sdk/src/SipClient.ts
```diff
@@ -123,7 +123,7 @@ export interface CreateSipParticipantOptions {
   hidePhoneNumber?: boolean;
   ringingTimeout?: number; // Duration in seconds
   maxCallDuration?: number; // Duration in seconds
-  enableKrisp?: boolean;
+  krispEnabled?: boolean;
```
lukasIO, line 126: this is exposed to users right now,  right? 
Vould we keep both, deprecate the old one and apply whichever of the two is set?

davidzhao: this is a relatively new, and still undocumented feature. so while it's a "breaking" change, I felt that the impact is low (since undocumented/unannounced), and relatively easy to change on the user's end

lukasIO: makes sense

lukasIO (approved):

PR #420, opened 2025-02-19
Fix utf8 split operation

lukasIO:
see https://github.com/livekit/client-sdk-js/pull/1414

### DIFF
--- a/packages/livekit-rtc/src/participant.ts
+++ b/packages/livekit-rtc/src/participant.ts
@@ -301,13 +301,13 @@ export class LocalParticipant extends Participant {
     const writableStream = new WritableStream<string>({
       // Implement the sink
       async write(text) {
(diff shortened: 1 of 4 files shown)


packages/livekit-rtc/src/participant.ts
```diff
@@ -301,13 +301,13 @@ export class LocalParticipant extends Participant {
     const writableStream = new WritableStream<string>({
       // Implement the sink
       async write(text) {
-        for (const textChunk in splitUtf8(text, STREAM_CHUNK_SIZE)) {
+        for (const textByteChunk of splitUtf8(text, STREAM_CHUNK_SIZE)) {
```
lukasIO, line 304: the `in` vs `of` bug here is the same as https://github.com/livekit/client-sdk-js/pull/1412

davidzhao (approved):

PR #422, opened 2025-02-19
update builder images

matkam:
updating images ahead of deprecation. updating all images, not just ubuntu-20.04.
https://github.com/actions/runner-images/issues/11101


theomonnom (approved):

lukasIO:
unfortunately this broke the ci runs for macos, we've gehst issues with the macos images recently and switched to macos-15 images that worked

PR #427, opened 2025-02-24
expose sipNumber in SipClient.CreateSIPParticipant

davidzhao:
also passes along missing participantAttributes

### DIFF
--- a/packages/livekit-server-sdk/src/SipClient.ts
+++ b/packages/livekit-server-sdk/src/SipClient.ts
@@ -108,10 +108,18 @@ export interface CreateSipDispatchRuleOptions {
 }
 
 export interface CreateSipParticipantOptions {
+  // Optional SIP From number to use. If empty, trunk number is used.
+  fromNumber?: string;
+  // Optional identity of the SIP participant
   participantIdentity?: string;
+  // Optional name of the participant
   participantName?: string;
+  // Optional metadata to attach to the participant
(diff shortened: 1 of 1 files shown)


packages/livekit-server-sdk/src/SipClient.ts
```diff
@@ -108,10 +108,18 @@ export interface CreateSipDispatchRuleOptions {
...
+  // Optional identity of the SIP participant
   participantIdentity?: string;
+  // Optional name of the participant
   participantName?: string;
+  // Optional metadata to attach to the participant
   participantMetadata?: string;
+  // Optional attributes to attach to the participant
   participantAttributes?: { [key: string]: string };
+  // Optionally send following DTMF digits (extension codes) when making a call.
+  // Character 'w' can be used to add a 0.5 sec delay.
```
lukasIO, line 122: is that some sort of standard? first time I'm hearing this

dennwc: It's not a standard, but this is how Twilio does it too

lukasIO (approved):

packages/livekit-server-sdk/src/SipClient.ts
davidzhao: I believe so. @dennwc could confirm

I took this from the protobuf docs, since the layer of translation obfuscated it

PR #428, opened 2025-02-24
fix: pass through correct disconnect reason

davidzhao:


packages/livekit-rtc/src/room.ts
```diff
@@ -249,8 +247,11 @@ export class Room extends (EventEmitter as new () => TypedEmitter<RoomCallbacks>
       this.emit(RoomEvent.ParticipantConnected, participant);
     } else if (ev.case == 'participantDisconnected') {
       const participant = this.remoteParticipants.get(ev.value.participantIdentity!);
-      this.remoteParticipants.delete(participant!.identity!);
-      this.emit(RoomEvent.ParticipantDisconnected, participant!);
+      if (participant) {
```
theomonnom, line 250: This condition isn't necessary, if the Rust SDK is wrong, then catching issues here is too late anyway. It may obfuscate debugging

theomonnom (approved):
otherwise lg

packages/livekit-rtc/src/room.ts
davidzhao: fair enough.. I'm just not a fan of seeing `!` in ts. and favor defensive programming regarding types. i've reverted this for now since we use this pattern everywhere.

PR #433, opened 2025-03-10
feat(rtc): add musl builds

nbsp:
tested locally on Alpine Linux edge, but currently lacks CI.


theomonnom:
mmh this is technically not possible with glibc?

PR #440, opened 2025-03-19
Add ForwardParticipant method to room service

cnderrauber:


lukasIO (approved):

packages/livekit-server-sdk/package.json
```diff
@@ -43,7 +43,7 @@
   },
   "dependencies": {
     "@bufbuild/protobuf": "^1.7.2",
-    "@livekit/protocol": "^1.32.1",
+    "@livekit/protocol": "^1.35.0",
```
lukasIO, line 46: we should probably bump the protocol here again to have the latest changes around KindDetails in there

PR #442, opened 2025-03-25
[feat:webhooks] Add support for customizing JWT clock tolerance for webhook verification

Wundero:
This PR adds a parameter to webhook reception (and access token verification) to support Jose's clock tolerance parameter for environments which may be slow or have other clock skew issues.


davidzhao (approved):
lgtm

PR #454, opened 2025-04-01
Clean up FFI handle when streaming VideoFrames

Shubhrakanti:
We need to clean up the FFIHandle after we copy over the frame data intro Node.

Memory usage before fix - `1700MB` after 2.5 mins,
Memory usage after fix - `~310MB` after 2.5 mins.

### DIFF
--- a/.changeset/violet-actors-join.md
+++ b/.changeset/violet-actors-join.md
@@ -1,0 +1,5 @@
+---
+'@livekit/rtc-node': patch
+---
+
+Fix memory leak with FFI Handle on VideoFrame creation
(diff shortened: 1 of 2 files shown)


Shubhrakanti:
Python equivalent https://github.com/livekit/python-sdks/blob/9e63c14dd378111719d4665d791a847c68e662be/livekit-rtc/livekit/rtc/video_frame.py#L108

Shubhrakanti:
@theomonnom in python [we rely on the `__del__` method](https://github.com/livekit/python-sdks/blob/9e63c14dd378111719d4665d791a847c68e662be/livekit-rtc/livekit/rtc/_ffi_client.py#L95) being called by the GC to clean up the handle. Here we do it explicitly by calling `dispose()` - I think that should be okay?

theomonnom (approved):

davidzhao (approved):

PR #457, opened 2025-04-02
Audio filter support

typester:


packages/livekit-rtc/src/audio_stream.ts
```diff
@@ -24,17 +29,41 @@ export class AudioStream implements AsyncIterableIterator<AudioFrame> {
...
+  constructor(track: Track, sampleRate: number);
+  constructor(track: Track, sampleRate: number, numChannels: number);
+  constructor(track: Track, noiseCancellationOptions: NoiseCancellationOptions);
+
+  constructor(
+    track: Track,
+    sampleRateOrNcOptions?: number | NoiseCancellationOptions,
+    numChannels?: number,
+    ncOptions?: NoiseCancellationOptions,
+  ) {
```
typester, line 44: This change is a bit tricky, but it was designed to allow `NoiseCancellationOptions` to be passed in a simpler way while maintaining backward compatibility.

It now allows NoiseCancellationOptions to be passed directly to AudioStream, eliminating the need to specify separate sampling rate and channel parameters.

The rationale is that when configuring audio filters, customizing the sampling rate or number of channels is rarely necessary, as our webrtc binding internally uses a 48kHz sampling rate. Using different values would require resampling, which can degrade performance.

The new constructor signature offers a more intuitive interface that aligns with common use cases—developers typically want to enable noise cancellation using default audio parameters, without having to manually set technical details that are best left at their optimal defaults.

bcherry (approved):

packages/livekit-rtc/src/audio_filter.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    const res = FfiClient.instance.request<LoadAudioFilterPluginResponse>({
+      message: {
+        case: 'loadAudioFilterPlugin',
+        value: req,
+      },
+    });
+
+    if (res.error) {
+      throw new Error(`Failed to initialize audio filter: ${res.error}`);
```
lukasIO, line 26: not sure if I'm a fan of the constructor throwing an error 🤔 
are we using the same pattern in python?

packages/livekit-rtc/src/audio_stream.ts
lukasIO: cool, this reasoning makes a lot of sense to me and nice work on handling all those overloads - always a pain. 

I'm wondering if we can make our (future) lives a bit easier if we add an overload in the form or 
```ts
constructor(track: Track, options: AudioStreamOptions);
```

where `AudioStreamOptions` could incorporate both, `NoiseCancellationOptions` and existing constructor params, something like
```ts
interface AudioStreamOptions: {
  noiseCancellation?: NoiseCancellationOptions,
  sampleRate?: number,
  numChannels?: number,
```

this would allow us to easily add additional fields in the future without additional overload management

typester: That sounds fair! I'll update it.

Just to confirm: since we want to maintain backward compatibility, we’d still use an overload for this, right? So basically, we’d replace the new one I added that takes NoiseCancellationOptions with one that takes AudioStreamOptions instead?

packages/livekit-rtc/src/audio_filter.ts
typester: Yeah, we’re doing it that way in Python as well. But I don't have any strong opinion for this.
Do you think it’d be better to add a separate function like `load` or something like that and handle errors there instead?

That said, this is currently only used internally by the Krisp filter, so it’s not something regular users would directly interact with.

lukasIO: if we're doing it in python already then let's stick to that pattern! 
I'm also not totally against it just feels a little weird. The fact that it's used only internally is good context, even less concerns from my side then

packages/livekit-rtc/src/audio_stream.ts
typester: Added `AudioStreamOptions` in [ea3e1ec](https://github.com/livekit/node-sdks/pull/457/commits/ea3e1ec63ae7b7a7ccf589ab90e0ebcf826d53e7)!

lukasIO (approved):
looks good to me! 🚀

PR #458, opened 2025-04-04
Use node stream for AudioStream and VideoStream

Shubhrakanti:
Use node streams to buffer incoming audio and video frames


packages/livekit-rtc/src/utils.ts
```diff
@@ -40,3 +40,50 @@ export function splitUtf8(s: string, n: number): Uint8Array[] {
...
+
+  constructor(private readonly capacity: number = 0) {
+    this.resolve = null;
+    this.capacity = capacity;
+  }
+
+  /**
+   * Push an item into the buffer. If the buffer is full, the newest item will
+   * be removed.
+   */
```
Shubhrakanti, line 64: This is what the [python implementation](https://github.com/livekit/python-sdks/blob/bee5b2f329b3bd462003c45c9756ceda3fb5345e/livekit-rtc/livekit/rtc/_utils.py#L50-L66) does. Wouldn't it make sense if the oldest time is evicted?

packages/livekit-rtc/src/utils.ts
```diff
@@ -40,3 +40,50 @@ export function splitUtf8(s: string, n: number): Uint8Array[] {
...
+      }
+      this.buffer.push(item);
+    }
+  }
+
+  /**
+   * Get an item from the buffer. If the buffer is empty, this will return a promise
+   * that resolves when a new item is pushed.
+   */
+  async get(): Promise<T> {
```
Shubhrakanti, line 81: I think it's best to just move the locking inside here. Are there any downsides to this?

lukasIO (commented):
I've talked to @nbsp in the past about moving more into the direction of utilising built in node (web) streams for our queues. 

See https://github.com/livekit/agents-js/pull/194/files for a first step into that direction. 

I'm not entirely sure it's the right call for all our queuing logic, but in general, if there's a way to make this work consistently across the repo with built-in primitives that's what I'd prefer.

Having custom implementations of both `RingQueue` and `AsyncIterableQueue` sounds a bit redundant though (esp. when the `RingQueue` in this PR acts as an async queue and we don't really make use of the capacity parameter. 
If we stick to a custom implementation, let's think about merging these two into one.

We've had some bugs in the past around async queuing and having custom implementations adds one more place where things can potentially go wrong.

Shubhrakanti:
> I've talked to @nbsp in the past about moving more into the direction of utilising built in node (web) streams for our queues.
> 
> See https://github.com/livekit/agents-js/pull/194/files for a first step into that direction.

Thanks super helpful context.

> I'm not entirely sure it's the right call for all our queuing logic, but in general, if there's a way to make this work consistently across the repo with built-in primitives that's what I'd prefer.
> 
> Having custom implementations of both `RingQueue` and `AsyncIterableQueue` sounds a bit redundant though, esp. when the `RingQueue` in this PR acts as an async queue and we don't really make use of the capacity parameter. If we stick to a custom implementation, let's think about merging these two into one.

This makes total sense let me see if I can get this working with `node:stream/web`.

packages/livekit-rtc/src/audio_stream.ts
```diff
@@ -47,54 +53,79 @@ export class AudioStream implements AsyncIterableIterator<AudioFrame> {
...
+        this.onEvent = (ev: FfiEvent) => {
+          if (
+            ev.message.case === 'audioStreamEvent' &&
+            ev.message.value.streamHandle === this.ffiHandle.handle
+          ) {
+            if (controller.desiredSize && controller.desiredSize > 0) {
+              controller.enqueue(ev);
+            } else {
+              console.warn('Audio stream buffer is full, dropping frame');
+            }
```
Shubhrakanti, line 67: [As I mentioned earlier]

(https://github.com/livekit/node-sdks/pull/458#discussion_r2028110185), the [python implementation]

> This is what the [python implementation](https://github.com/livekit/python-sdks/blob/bee5b2f329b3bd462003c45c9756ceda3fb5345e/livekit-rtc/livekit/rtc/_utils.py#L50-L66) does. Wouldn't it make sense if the oldest time is evicted?


@theomonnom if this needs to be the behavior here as well then maybe node streams are not a good choice because we don't have explicit control over the buffer eviction policy.

packages/livekit-rtc/src/audio_stream.ts
```diff
@@ -47,54 +53,81 @@ export class AudioStream implements AsyncIterableIterator<AudioFrame> {
...
-        } else {
-          this.eventQueue.push(frame);
+    const source = new ReadableStream<FfiEvent>({
+      start: (controller) => {
+        this.onEvent = (ev: FfiEvent) => {
+          if (
+            ev.message.case === 'audioStreamEvent' &&
+            ev.message.value.streamHandle === this.ffiHandle.handle
+          ) {
+            if (infinite_capacity || (controller.desiredSize && controller.desiredSize > 0)) {
```
Shubhrakanti, line 65: This is kind of a hack to make this stream have an infinite capacity buffer. Makes me wonder if this is not the correct interface.

packages/livekit-rtc/src/audio_stream.ts
```diff
@@ -47,54 +53,81 @@ export class AudioStream implements AsyncIterableIterator<AudioFrame> {
...
+              break;
+          }
+        },
+      },
+      {
+        highWaterMark: capacity > 0 ? capacity : undefined,
+      },
+    );
+
+    this.reader = source.pipeThrough(transformStream).getReader();
```
lukasIO, line 113: we could acquire the reader lazily in `next()` and also make sure to call `releaseLock` when done

packages/livekit-rtc/src/audio_stream.ts
```diff
@@ -47,54 +53,81 @@ export class AudioStream implements AsyncIterableIterator<AudioFrame> {
...
-        break;
-      case 'eos':
-        FfiClient.instance.off(FfiClientEvent.FfiEvent, this.onEvent);
-        break;
-    }
-  };
+      },
+    });
+
+    const transformStream = new TransformStream<FfiEvent, AudioFrame>(
```
lukasIO, line 82: is the transform actually needed? wondering if we can handle this directly in the ReadableStream?

packages/livekit-rtc/src/video_stream.ts
```diff
@@ -12,30 +11,22 @@ import type {
...
-  /** @internal */
-  queueResolve: ((value: IteratorResult<VideoFrameEvent>) => void) | null = null;
-  /** @internal */
-  mutex = new Mutex();
 
-  track: Track;
+class VideoStreamSource implements UnderlyingSource<VideoFrameEvent> {
+  private controller?: ReadableStreamDefaultController<VideoFrameEvent>;
+  private ffiHandle: FfiHandle;
+  private closed = false;
```
Shubhrakanti, line 26: technically making these private would not be backward compatible.

packages/livekit-rtc/src/video_stream.ts
```diff
@@ -69,54 +62,39 @@ export class VideoStream implements AsyncIterableIterator<VideoFrameEvent> {
...
-            rotation: value.rotation!,
-          });
+        const videoFrameEvent = {
+          frame: value.frame,
+          timestampUs: value.timestampUs!,
+          rotation: value.rotation!,
+        };
+
+        if (!this.closed) {
+          this.controller.enqueue(videoFrameEvent);
```
Shubhrakanti, line 72: https://github.com/livekit/node-sdks/pull/458#discussion_r2036696465

This is okay to do. The official standard [has an example](https://github.com/livekit/node-sdks/pull/458#discussion_r2036696465) of a push stream with no back pressure support. We should just make a comment making it explicit. If the request arises we should limit the queue size based on memory size in bytes not number of frames.

Shubhrakanti (commented):
### TLDR 
We can't have `Video/AudioStream` be a `ReadableStream` without introducing breaking changes to the current api. Right now we don't automatically close the stream when an iterator finishes. If we extend the `ReadableStream` interface, this behavior is unavoidable. 

### Explanation
`AsyncIterators` in JS have a [`return` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#returnvalue_2) that is invoked when iterator stops or exits due to a `break;` statement. When this happens, the `ReadableStream` under the hood will invoke the `cancel()` method on the source. 

For example if we had code like this 
```typescript
const stream = new VideoStream(track);
    let i = 0;
    for await (const frame of stream) {
        i++;
        console.log('loop 1 timestamp', frame.timestampUs);
        if (i > 5) {
            break;
        }
    }
    console.log('finished loop 1');
    for await (const frame of stream) {
      i++;
      console.log('loop 2 timestamp', frame.timestampUs);
      if (i > 10) {
          break;
      }
    }
    console.log('finished loop 2');
   }
```

**Output from existing implementation.** 
```
loop 1 timestamp 352996437000n
loop 1 timestamp 352996470000n
loop 1 timestamp 352996527000n
...
finished loop 1 -> no resource clean up
loop 2 timestamp 352996537000n
loop 2 timestamp 352996660000n
loop 2 timestamp 352996957000n
...
finished loop 2
```
Loop 2 pulls of the stream because there is no resource clean up.

**Output from new implementation with `ReadableStream`**
```
loop 1 timestamp 352996437000n
loop 1 timestamp 352996470000n
loop 1 timestamp 352996527000n
...
finished loop 1 -> under the hood source.close() is called
finished loop 2
```
Loop 2 is empty because the stream is closed.



I _tried_ to hack around this by doing something like 
```typescript
  cancel(reason?: any) {
    // do nothing
  }

  close() {
    this.closed = true;
    FfiClient.instance.off(FfiClientEvent.FfiEvent, this.onEvent);
    this.ffiHandle.dispose();
    console.log('cancelled');
  }
}

export class VideoStream extends ReadableStream<VideoFrameEvent> {
  constructor(track: Track) {
    this.source = new VideoStreamSource(track);
    super(this.source);
  }

  close() {
    this.source.close();
  }
}
```

but that doesn't work because you use `this` before`super`. See [this post](https://stackoverflow.com/questions/57812128/what-is-the-best-way-to-define-logic-for-web-streams-sources-sinks) for more info.


### Proposals

**1. Make the breaking change and update the API with a new version.**

I think this is the better solution because we get all the benefits of `ReadableStream` (like locking, `tee()`, `pipeTo/Through`). 

I also think the current implementation allows user code to have subtle bugs. For example above, if we do something like 

```typescript
for await (const frame of stream) {
  ..
}

await someReallyLongRunningFunction()

for await (const frame of stream) {
  ..
}

```

The `FFiHandle` in the `VideoStream` will continue to push frames into the buffer even though we're not consuming from the stream.

**2. Just encapsulate `VideoStream` with  a 

This makes the current implementation logic cleaner and there are no breaking changes, but we loose the benefits of `ReadableStream`.

lukasIO (commented):
this looks very clean! 

I think you might have accidentally removed the AudioStream implementation?

packages/livekit-rtc/src/video_stream.ts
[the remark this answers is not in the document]
Shubhrakanti: Discussed offline. It's okay

packages/livekit-rtc/src/audio_stream.ts
```diff
@@ -95,43 +82,31 @@ export class AudioStream implements AsyncIterableIterator<AudioFrame> {
...
+  cancel() {
+    FfiClient.instance.off(FfiClientEvent.FfiEvent, this.onEvent);
     this.ffiHandle.dispose();
   }
+}
 
-  [Symbol.asyncIterator](): AudioStream {
-    return this;
+export class AudioStream extends ReadableStream<AudioFrame> {
+  constructor(
```
lukasIO, line 105: we would want to add the more specific overload constructors here that @typester had added for the AudioStream

Shubhrakanti: oh right. Whoops.

PR #466, opened 2025-04-17
Debug teeing node stream

Shubhrakanti:
Run with `cd packages/livekit-rtc && pnpm build && cd ../../examples/monitor-memory-usage && pnpm tsx index.ts`

### DIFF
--- a/examples/monitor-memory-usage/index.ts
+++ b/examples/monitor-memory-usage/index.ts
@@ -1,0 +1,79 @@
+import { VideoStream, Room, RoomEvent, TrackKind } from '@livekit/rtc-node';
+import { config } from 'dotenv';
+import * as fs from 'fs';
+import { AccessToken } from 'livekit-server-sdk';
+
+config();
+
+// create access token from API credentials
+const token = new AccessToken(process.env.LIVEKIT_API_KEY, process.env.LIVEKIT_API_SECRET, {
+  identity: 'node-agent',
+});
+token.addGrant({
+  room: 'test-room',
+  roomJoin: true,
+  roomCreate: true,
+  canPublish: true,
+  canPublishData: true,
+});
+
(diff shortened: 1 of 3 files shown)


Shubhrakanti:
What's funny is if I of 

```typescript
const [stream1, stream2] = stream.tee();
console.log('Created tee branches');

// Use getReader() instead of for-await
const reader1 = stream1.getReader();
let i = 0;

try {
  while (true) {
    const {value, done} = await reader1.read();
    if (done) break;
    
    i++;
    console.log('loop 1 timestamp', value.timestampUs);
    if (i > 5) {
      console.log('Breaking out of loop 1');
      break;
    }
  }
} catch (err) {
  console.error('Error in loop 1:', err);
} finally {
  // Explicitly release the lock on the reader
  reader1.releaseLock();
}
console.log('finished loop 1');

// Now read from the second branch
const reader2 = stream2.getReader();
try {
  while (true) {
    const {value, done} = await reader2.read();
    if (done) break;
    
    i++;
    console.log('loop 2 timestamp', value.timestampUs);
    if (i > 10) {
      console.log('Breaking out of loop 2');
      break;
    }
  }
} catch (err) {
  console.error('Error in loop 2:', err);
} finally {
  reader2.releaseLock();
}
console.log('finished loop 2');
```
it works fine, but ideally we can get this to work with `for .. await`

lukasIO:
[this comment](https://github.com/whatwg/streams/issues/1255#issuecomment-1421672493) has some insight on this, as the spec for readable streams says the underlying stream closes if you break out of the for ... await loop. 

I'm not sure why we don't even see the log `'finished loop 1'` though. That seems odd. 

What makes it work is if you loop over it like this
```ts
    for await (const frame of stream1.values({ preventCancel: true })) {
```

so I believe your implemenation is still correct, it's just that the spec says the stream will close once you break out of the loop.

PR #486, opened 2025-06-02
Fix memory leak of AudioSource

typester:
It's actually an issue on Track side where it keep a reference for AudioSource and needs to be disposed

### DIFF
--- a/.changeset/selfish-moons-tie.md
+++ b/.changeset/selfish-moons-tie.md
@@ -1,0 +1,6 @@
+---
+"example-publish-wav": patch
+"@livekit/rtc-node": patch
+---
+
+Fix memory leak of AudioSource


typester:
I'm going to apply this fix to agent-js after this fix is released. Also I will do the documentation improvement on that.

Shubhrakanti (approved):
This makes sense to me.

packages/livekit-rtc/src/track.ts
```diff
@@ -45,6 +45,10 @@ export abstract class Track {
   get muted(): boolean | undefined {
     return this.info?.muted;
   }
+
+  async close() {
```
lukasIO, line 49: should we have a optional boolean here to also close the source when closing the track? e.g. 
`async close(closeSource = true) {`

typester: @Shubhrakanti @bcherry I'm not familiar with JS side but this sounds fair to me. What do you think?

I'm going to merge this pr but I'm open to implement this in a different pr

Shubhrakanti: This makes sense to me!

typester: Thanks, I will create an other pr for it!

PR #487, opened 2025-06-04
Remove all ReadableStream Imports

toubatbrian:
Upgrade `@type/node` and remove ReadableStream in favor of global ReadableStream

### DIFF
--- a/packages/livekit-rtc/src/participant.ts
+++ b/packages/livekit-rtc/src/participant.ts
@@ -20,21 +20,29 @@ import {
   ParticipantKind,
 } from './proto/participant_pb.js';
 import {
+  ChatMessage as ChatMessageModel,
   DataStream_ByteHeader,
   DataStream_Chunk,
   DataStream_Header,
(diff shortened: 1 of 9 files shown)


packages/livekit-rtc/src/participant.ts
```diff
@@ -460,7 +458,7 @@ export class LocalParticipant extends Participant {
     const fileStats = await stat(path);
     const file = await open(path);
     try {
-      const stream: ReadableStream<Uint8Array> = file.readableWebStream({ type: 'bytes' });
```
toubatbrian, line 463: Remove the payload since `file.readableWebStream` drops the argument since node.js v22
Screenshot 2025-06-04 at 12 14 01 AM

Screenshot 2025-06-04 at 12 15 25 AM

Shubhrakanti (approved):
Would be good to test the data streams manually if you can. See if there is an example that does this for you.

lukasIO:
@toubatbrian does this change still work in v20?

PR #489, opened 2025-06-06
Add support for destination_country in outbound trunk

biglittlebigben:


packages/livekit-server-sdk/src/SipClient.ts
```diff
@@ -186,6 +187,17 @@ export interface SipTrunkUpdateOptions {
...
 
+export interface SipOutboundTrunkUpdateOptions {
+  numbers?: ListUpdate;
+  allowedAddresses?: ListUpdate;
+  allowedNumbers?: ListUpdate;
+  authUsername?: string;
+  authPassword?: string;
+  destinationCountry?: string;
+  name?: string;
+  metadata?: string;
```
davidzhao, line 198: what is metadata used for for outbound?

biglittlebigben: It is defined on the Outbound trunk, but I can't find any usage of it. Intent may have been to set a default value for the ParticipantMetadata, but I'm not sure.

PR #495, opened 2025-06-17
feat(livekit-rtc): Add `closeSource = true` default parameter to local tracks

typester:
Following up for #486


lukasIO (approved):
nice, I think this makes sense to simplify cleanup for the most common use cases

PR #498, opened 2025-06-23
do not mix changesets

davidzhao:
including private packages was causing changesets to fail


lukasIO (approved):
From their docs, it appears that the "correct" way of handling packages that should not be published is to simply set `private: true` in package.json and not add them to the ignored packages in the changesets config, [for reference ](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#ignore-array-of-packages)

PR #508, opened 2025-07-18
[Bugifx] Shubhra/ Add buffer to FFI events

Shubhrakanti:
This ports over a fix that was done in the python side, but never implemented  in the node sdk ([PR](https://github.com/livekit/python-sdks/pull/60)).

User reported that the `Room` would throw an error if there was already a participant inside and they tried to connect. This is because we were processing the `participant_connected` event before the `trackSubscribed` event. 

**Before fix:**
1. call `room.connect`
2. receive `trackSubscribed` ← Tries to find participant that was never created!
3. throw participant not found error

**After fix:**
1. call `room.connect`
2. receive `participants_updated`
3. receive ` participant_connected` ← **Creates the participant**
5. receive `track_published`
6. receive `track_subscribed` ← **Works because participant exists**

### DIFF
--- a/packages/livekit-rtc/src/ffi_client.ts
+++ b/packages/livekit-rtc/src/ffi_client.ts
@@ -4,6 +4,7 @@
 import type { PartialMessage } from '@bufbuild/protobuf';
 import type { TypedEventEmitter as TypedEmitter } from '@livekit/typed-emitter';
 import EventEmitter from 'events';
+import { log } from './log.js';
 import {
   FfiHandle,
   livekitCopyBuffer,
@@ -30,6 +31,49 @@ declare global {
   var _ffiClientInstance: FfiClient | undefined;
 }
 
+export class FfiQueue<T> {
+  private subscribers: Set<ReadableStreamDefaultController<T>> = new Set();
+
+  constructor() {}
+
+  put(item: T): void {
+    const toRemove: Set<ReadableStreamDefaultController<T>> = new Set();
+
+    for (const controller of this.subscribers) {
+      try {
+        controller.enqueue(item);
+      } catch (error: unknown) {
+        log.error(error, 'Error enqueuing item to stream');
+        toRemove.add(controller);
+      }
+    }
+
+    for (const controller of toRemove) {
+      this.subscribers.delete(controller);
+    }
+  }
+
+  subscribe(): ReadableStream<T> {
+    let controller: ReadableStreamDefaultController<T>;
+
+    const stream = new ReadableStream<T>({
+      start: (ctrl) => {
+        controller = ctrl;
+        this.subscribers.add(controller);
+      },
+      cancel: () => {
+        this.subscribers.delete(controller);
+      },
+    });
+
+    return stream;
+  }
+
+  unsubscribe(stream: ReadableStream<T>): void {
+    stream.cancel();
+  }
+}
+
 export class FfiClient extends (EventEmitter as new () => TypedEmitter<FfiClientCallbacks>) {
   /** @internal */
   static get instance(): FfiClient {
@@ -39,18 +83,25 @@ export class FfiClient extends (EventEmitter as new () => TypedEmitter<FfiClient
     return globalThis._ffiClientInstance;
   }
 
+  private _queue: FfiQueue<FfiEvent> = new FfiQueue<FfiEvent>();
   constructor() {
     super();
     this.setMaxListeners(0);
+    this._queue = new FfiQueue<FfiEvent>();
 
     livekitInitialize(
       (event_data: Uint8Array) => {
         const event = FfiEvent.fromBinary(event_data);
         this.emit(FfiClientEvent.FfiEvent, event);
+        this._queue.put(event);
       },
       true,
       SDK_VERSION,
     );
+  }
+
+  get queue(): FfiQueue<FfiEvent> {
+    return this._queue;
   }
 
   request<T>(req: PartialMessage<FfiRequest>): T {
--- a/packages/livekit-rtc/src/ffi_client.ts
+++ b/packages/livekit-rtc/src/ffi_client.ts
@@ -4,7 +4,6 @@
 import type { PartialMessage } from '@bufbuild/protobuf';
 import type { TypedEventEmitter as TypedEmitter } from '@livekit/typed-emitter';
 import EventEmitter from 'events';
-import { log } from './log.js';
 import {
   FfiHandle,
   livekitCopyBuffer,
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "lib": ["es2015"],
+    "lib": ["es2015", "ES2021.WeakRef"],
```
Shubhrakanti: We need this so that linter doesn't throw an error when using `FinalizationRegistry`

Shubhrakanti: @lukasIO is there a reason this is `es2015` can we just update it to something later.

packages/livekit-rtc/src/ffi_client.ts
```diff
@@ -30,6 +31,54 @@ declare global {
...
+        toRemove.add(controller);
+      }
+    }
+
+    for (const controller of toRemove) {
+      this.subscribers.delete(controller);
+    }
+  }
+
+  subscribe(): ReadableStream<T> {
```
Shubhrakanti: I chose this to be the equivalent of the asyncio's `Queue` we use in python. The main feature we care about is asynchronous reads and waiting for events while the stream is empty.

packages/livekit-rtc/src/room.ts
```diff
@@ -88,8 +90,19 @@ export class Room extends (EventEmitter as new () => TypedEmitter<RoomCallbacks>
   remoteParticipants: Map<string, RemoteParticipant> = new Map();
   localParticipant?: LocalParticipant;
 
+  private static cleanupRegistry = new FinalizationRegistry((cleanup: () => void) => {
```
Shubhrakanti: We use FinalizationRegistry ([see docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry)) to act as an equivalent of python's `__del__` method

Shubhrakanti: There's also Symbol.dispose as defined in a ECMA proposal https://github.com/tc39/proposal-explicit-resource-management but I wasn't sure if this was merged in yet.

packages/livekit-rtc/src/room.ts
```diff
@@ -237,10 +255,45 @@ export class Room extends (EventEmitter as new () => TypedEmitter<RoomCallbacks>
...
+          event.message.value?.message.case == 'eos'
+        ) {
+          break;
+        }
+        this.onFfiEvent(event);
+      }
+    } catch (error) {
+      log.debug(error, 'Listen task ended');
+    } finally {
+      this.listenTaskPromise = undefined;
```
Shubhrakanti: is it better to type it as `| null`?

packages/livekit-rtc/src/ffi_client.ts
```diff
@@ -30,6 +31,54 @@ declare global {
...
+        controller.enqueue(item);
+      } catch (error: unknown) {
+        log.error(error, 'Error enqueuing item to stream');
+        toRemove.add(controller);
+      }
+    }
+
+    for (const controller of toRemove) {
+      this.subscribers.delete(controller);
+    }
```
Shubhrakanti: is this overkill? Just want to make sure there are no memory leaks.

packages/livekit-rtc/src/ffi_client.ts
```diff
@@ -30,6 +31,54 @@ declare global {
...
+
+  /**
+   * @throws "TypeError: Invalid state: ReadableStream is locked"
+   * if the stream is locked, make sure the stream is not being read from
+   * before calling this method.
+   */
+  unsubscribe(stream: ReadableStream<T>): void {
+    stream.cancel();
+  }
+}
```
Shubhrakanti: I don't love this, but it works for now.

packages/livekit-rtc/src/room.ts
```diff
@@ -88,8 +90,19 @@ export class Room extends (EventEmitter as new () => TypedEmitter<RoomCallbacks>
   remoteParticipants: Map<string, RemoteParticipant> = new Map();
   localParticipant?: LocalParticipant;
 
+  private static cleanupRegistry = new FinalizationRegistry((cleanup: () => void) => {
+    cleanup();
+  });
   constructor() {
     super();
+    // Register a finalizer to disconnect the room when it's garbage collected
+    Room.cleanupRegistry.register(this, () => {
```
theomonnom: I don't think we need this. If the FfiHandle is dropped, the Rust side will already close the room

packages/livekit-rtc/src/room.ts
```diff
@@ -180,6 +193,9 @@ export class Room extends (EventEmitter as new () => TypedEmitter<RoomCallbacks>
       options,
     });
 
+    // subscribe before connecting so we don't miss any events
+    this.ffiQueue = FfiClient.instance.queue.subscribe();
```
theomonnom: On JS it may be simpler, just hooking into the FfiEvent using `on`. We could just append to a list?

theomonnom: Maybe it's fine, but IIRC that was the only difference. JS using events, but python using queues

packages/livekit-rtc/src/room.ts
Shubhrakanti: Good to know. I saw in the python side so that's why I implemented it - but had a felling we didn't need this.

packages/livekit-rtc/src/room.ts
Shubhrakanti: > On JS it may be simpler, just hooking into the FfiEvent using on that's exactly what this is doing. 

We grab the events and append them to a queue. The only reason it's a stream is so reads are async and we don't block the main event loop.

Shubhrakanti: Do you mean get rid of `FfiClient.instance.queue` all together. Replace this with 


```typescript

connect(){
// wire up the event listener first
FfiClient.instance.on(FfiClientEvent.FfiEvent, this.onFfiEvent)

// send the connection request
const res = FfiClient.instance.request<ConnectResponse>({ ...... })

}
//
onEvent(event): {
   this.buffer.push(event)
  // only start processing the events once we've connected.
   if (connected) {
      this.proccessEvent(event)
...
```

Shubhrakanti: As I understand it the order of operations are.

1. Create a buffer for all the events coming in
2. Connect to the room
3. Immediately populate the buffer so we don't drop events
4. Start processing events

Right now step 1 is done with `this.ffiQueue = FfiClient.instance.queue.subscribe();` if we replace it with `FfiClient.instance.on(FfiClientEvent.FfiEvent, this.onFfiEvent)` we still need to buffer the events to process until we receive the `ConnectResponse`?

tsconfig.json
Shubhrakanti: Looks like we don't need `FinalizationRegistry`. But still would be good to upgrade this?

Shubhrakanti:
Discussed offline with @theomonnom - we're just going to buffer the events that come in before the connect in `room.ts`. Once we have connected - we will process them in batch.

theomonnom (approved):

PR #509, opened 2025-07-19
Update AccessToken.ts

yaniv-peretz:
Not sure, but it's it required to allow users connect to a room (mine failed without it)


davidzhao:
optional is actually the right thing here. for room connect tokens, it's required, but for other types of tokens it isn't needed

Issue #515, opened 2025-08-07
Support `webhooks` in `EgressClient`


bgreenier:
### Describe the problem

I'd like to be able to take advantage of per-egress webhooks from the egress client - It looks like there's just a bit of plumbing missing to make this possible - The `rpc` request objects define `webhooks`, but the `startXXX` method options don't expose it.

### Describe the proposed solution

The ability to pass `webhooks` in `EgressClient::startXXX` method options.

### Alternatives considered

None

### Importance

nice to have

PR #516, opened 2025-08-07
feat(webhooks): Add webhooks to `EgressClient`

bgreenier:
Addresses https://github.com/livekit/node-sdks/issues/515

- Adds a `BaseOptions` for all egress-type options
- Passes along `webhooks` option to request objects
- Note: Does not add support to deprecated signatures


bgreenier:
Need to follow up with Gather legal before I can sign this CLA

davidzhao (approved):
nice work!

PR #526, opened 2025-08-19
Add sendBytes method

bcherry:


packages/livekit-rtc/src/participant.ts
```diff
@@ -481,6 +473,101 @@ export class LocalParticipant extends Participant {
...
+      // Treat as AsyncIterable (e.g., Node.js Readable of Uint8Array/Buffer)
+      for await (const chunk of data as AsyncIterable<Uint8Array>) {
+        await writer.write(chunk);
+        maybeReportProgress(chunk.byteLength);
+      }
+    } else {
+      throw new Error('Unsupported data type for sendBytes');
+    }
+
+    await writer.close();
```
lukasIO, line 565: this should be in a `finally` clause as you're throwing an error just above and the stream has already been opened

lukasIO (approved):
nice!

PR #527, opened 2025-08-26
Add docs generation workflows

rektdeckard:
https://linear.app/livekit/issue/DPRO-377/add-nodejs-reference-docs

- Sets up GitHub workflow docs deployment for both `server-sdk-js` and `rtc-node`

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -54,7 +54,7 @@ jobs:
       - name: Upload livekit-rtc docs to S3
         if: ${{ github.event.inputs.livekit-rtc }}
         run: |
-          aws s3 cp packages/livekit-rtc/docs/ s3://livekit-docs/client-sdk-node --recursive
+          aws s3 cp packages/livekit-rtc/docs/ s3://livekit-docs/rtc-node --recursive
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.DOCS_DEPLOY_AWS_ACCESS_KEY }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.DOCS_DEPLOY_AWS_API_SECRET }}


packages/livekit-server-sdk/src/crypto/digest.ts
```diff
@@ -9,6 +9,8 @@ export async function digest(data: string): Promise<ArrayBuffer> {
     return crypto.subtle.digest('SHA-256', encoder.encode(data));
   } else {
     const nodeCrypto = await import('node:crypto');
-    return nodeCrypto.createHash('sha256').update(data).digest();
+    const buf = nodeCrypto.createHash('sha256').update(data).digest();
+    // NOTE: we know this will always be an ArrayBuffer, not a SharedArrayBuffer
```
lukasIO: how do we know this? 👀

.changeset/eleven-lions-tan.md
```diff
@@ -0,0 +1,8 @@
+---
+"@livekit/rtc-node": patch
+"livekit-server-sdk": patch
+---
+
+Add docs generation workflows
+Fix error logging format strings in `rtc-node`
```
lukasIO: in general each user facing change should get its own changeset file, so each line you've written here would be its own file, e.g. with the `ArrayBuffer` change only causing a `patch` for rtc-node

packages/livekit-rtc/rust-sdks
lukasIO, line 1: was this needed or can we revert this?

packages/livekit-rtc/src/data_streams/stream_reader.ts
```diff
@@ -58,7 +58,7 @@ export class ByteStreamReader extends BaseStreamReader<ByteStreamInfo> {
             this.handleChunkReceived(value);
             return { done: false, value: value.content! };
           }
-        } catch (error) {
+        } catch (error: any) {
           log.error('error processing stream update', error);
```
lukasIO: similar to your findings above, we should probably treat the `error` here as the first argument (merging object) ?

bcherry (commented):
is the rust sdks update intentional? there's also a crypto update that seems unlikely to be related to docs.

lukasIO:
@rektdeckard since last week most CI workflows were broken. I extracted the bare minimum of necessary fixes in https://github.com/livekit/node-sdks/pull/529. 
This should make it easier to reduce the necessary changes in this PR

packages/livekit-server-sdk/src/crypto/digest.ts
rektdeckard: This is a Node Typescript artifact. The `slice()` method always represents the same type ([web](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice), [node](https://nodejs.org/api/buffer.html#bufbuffer)) as the underlying concrete `ArrayBufferLike`, and the `Buffer.buffer` property is always an `ArrayBuffer`.

.changeset/eleven-lions-tan.md
rektdeckard: Copy! I'm going to pull out the non-docgen related patches into separate PRs.

packages/livekit-rtc/rust-sdks
rektdeckard: We can revert, I had just never cloned this repo before and syncing submodules pulled in the latest.

packages/livekit-rtc/src/data_streams/stream_reader.ts
rektdeckard: Removing, I see you've already made the change here: https://github.com/livekit/node-sdks/pull/529/files#diff-de8971bb9a5b39bcb8d8653fcb273b9eff4b34fc07355501429f89462072a96bL61

packages/livekit-server-sdk/src/crypto/digest.ts
rektdeckard: Moving to its own PR

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,70 @@
+# SPDX-FileCopyrightText: 2025 LiveKit, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: Publish Docs
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
```
lukasIO, line 10: this looks like triggering publishing of docs is exclusively a manual process. 

It would be great if we could automate it as part of the release workflow, i.e. if a package got published, also publish the latest docs, so that the docs are always up to date with the latest release without us having to remember explicitly publishing the docs.

PR #529, opened 2025-08-28
Fix broken builds

lukasIO:
will follow up with a house keeping PR to address 
- different TS versions used across the monorepo
- different protobuf versions
- build tests not running reliably on PRs


boks1971 (commented):
lgtm!

PR #536, opened 2025-09-22
Shubhra/log room event with env var

Shubhrakanti:
Makes debugging easier instead of having to edit `node_modules` from the agents directory


Shubhrakanti:
Just manually tested with an example this works okay.

Issue #538, opened 2025-09-22
negative currentQueueSize


matej-dataqueue:
### Select which package(s) are affected

\@livekit/rtc-node

### Describe the bug

The `currentQueueSize` variable in AudioSource can be negative. 

### What I'm expecting
`currentQueueSize` should never be negative, as implied by the name. 

### What happens
`currentQueueSize` is allowed to be negative. The getter returns 0, but internally, it is negative. 

### Reproduction

Code taken from `AudioSource.ts` `captureFrame()`: 

```typescript
const now = Number(process.hrtime.bigint() / BigInt(1000000));
const elapsed = this.lastCapture === 0 ? 0 : now - this.lastCapture;
const frameDurationMs = (frame.samplesPerChannel / frame.sampleRate) * 1000;
this.currentQueueSize += frameDurationMs - elapsed;
```

If `elapsed > frameDurationMs`, we are subtracting from `currentQueueSize` here. That makes sense. 
However, if `elapsed > frameDurationMs + currentQueueSize`, we are allowing `currentQueueSize` to be negative. 

Example:

I send 1000ms of audio. 
`currentQueueSize = 1000`
I wait 5 seconds. 
I send 1000ms of audio. 
`currentQueueSize = -3000`
I send 1000ms of audio. 
`currentQueueSize = -2000` => `queuedDuration = 0` **even though there is 2000ms of audio queued!**

### Logs

```shell
not relevant
```

### System Info

```shell
not relevant
```

### LiveKit server version

LiveKit cloud

### Severity

serious, but I can work around it

### Additional Information

Suggested fix:

```typescript
this.currentQueueSize = frameDurationMs + Math.max(this.currentQueueSize - elapsed, 0);
```

PR #539, opened 2025-09-22
Fix negative queue size

matej-dataqueue:
https://github.com/livekit/node-sdks/issues/538


lukasIO (approved):
thanks for the fix, this makes sense!

PR #544, opened 2025-09-25
Adding display name to CreateSipParticipant

alexlivekit:


pnpm-lock.yaml
```diff
@@ -995,6 +995,9 @@ packages:
   '@livekit/protocol@1.40.0':
     resolution: {integrity: sha512-1q0TNqlSTDW9ZuQCYTLDusyO+StvAXPmECQgyuszThDjzhTwQOkA6Rv7B1wPcvpTcwieIG0vuAO4cw4+kg+xWA==}
 
+  '@livekit/protocol@1.42.0':
```
bcherry, line 992: I noticed that the lockfile now has two versions of protocol specified. it seems that's because examples/agent-dispatch/package.json has its own specific dependency but it's not needed. you should remove the explicit protocol dependency from that example's package.json, add it to the main package.json, then rerun `pnpm install`.

package.json
```diff
@@ -18,6 +18,9 @@
     "test": "vitest run",
     "test:watch": "vitest"
   },
+  "dependencies": {
+    "@livekit/protocol": "^1.42.0"
+  },
```
bcherry: you should be able to remove this one

Issue #472, opened 2025-05-07
undefined participant.trackPublications


tobowers:
### Select which package(s) are affected

\@livekit/rtc-node

### Describe the bug

Uncaught exception detected: undefined is not an object (evaluating 'participant.trackPublications'

We see this sometimes (actually quite often) in our `@livekit/node-rtc` server side. We are waiting for the remote participant, and it *joins* (we see it join in the web client), but this error is triggered and we never see it in the remoteParticipants (even 10s later or so). It's triggered a few times in a row.

### Reproduction

not sure how to reliably do this, but it happens very often where are :

Creating a new room, having node-rtc client connect *and* the unity client connect. The error is happening below our code.

### Logs

```shell
error: Uncaught exception detected: undefined is not an object (evaluating 'participant.trackPublications') {"alert":true,"error":{"message":"undefined is not an object (evaluating 'participant.trackPublications')"},"errorCount":1,"timestamp":"2025-05-07T09:11:50.704Z"}
```

### System Info

```shell
bun 1.2.9
rtc-node 0.13.0
```

### LiveKit server version

LiveKit cloud

### Severity

serious, but I can work around it

tobowers:
tracked it down to this line: https://github.com/livekit/node-sdks/blob/266add86514122317ab362a881a25a737794cad8/packages/livekit-rtc/src/room.ts#L267

pzerelles:
I tracked down a similar problem, or maybe it is the same, where the response for a `publishTrack` event is awaited as a promise and the event loop processes other events first, e.g. the `localTrackPublished` event, that requires a track to be in the local participants `trackPublications` map, but because of different handling of the event loop in Bun, this leads to an error. 

I will create a PR with a patch that will force the event loop to finish outstanding promises first before the `localTrackPublished` event is processed.

PR #546, opened 2025-10-01
fix: delay processing of localTrackPublished event because of race condition

### LINKED ISSUE #462: Race condition for local track subscribed
### Select which package(s) are affected

\@livekit/rtc-node

### Describe the bug

race condition in track subscription.

We have some code that that kept crashing our server yesterday. There's a problem with this line:
https://github.com/livekit/node-sdks/blob/f19824689df51ed1e954e2da9b57f04eb801e914/packages/livekit-rtc/src/room.ts#L262

Where sometimes (it's really unclear when) the publication can be undefined:
```
TypeError#1: undefined is not an object (evaluating 'publication.resolveFirstSubscription')
```

You can see on that line that there's a ! in the typescript, so lint must have seen it.

This happened when we were reconnecting to a room with same participant name, same room name. So maybe the track was still published from the previous connection?

### Reproduction

unfortunately not sure

### Logs

```shell

```

### System Info

```shell
Bun 1.2.8 
"@livekit/rtc-node": "^0.13.10",
"livekit-server-sdk": "^2.11.0",
```

### LiveKit server version

LiveKit cloud

### Severity

annoyance

### Additional Information

_No response_

pzerelles:
Maybe fixes #462 and #472. 

Without this, I cannot start an agent using Bun runtime.

### DIFF
--- a/packages/livekit-rtc/src/room.ts
+++ b/packages/livekit-rtc/src/room.ts
@@ -334,8 +334,10 @@ export class Room extends (EventEmitter as new () => TypedEmitter<RoomCallbacks>
       participant!.info.disconnectReason = ev.value.disconnectReason;
       this.emit(RoomEvent.ParticipantDisconnected, participant!);
     } else if (ev.case == 'localTrackPublished') {
-      const publication = this.localParticipant!.trackPublications.get(ev.value.trackSid!);
-      this.emit(RoomEvent.LocalTrackPublished, publication!, this.localParticipant!);
+      setImmediate(() => {
+        const publication = this.localParticipant!.trackPublications.get(ev.value.trackSid!);
+        this.emit(RoomEvent.LocalTrackPublished, publication!, this.localParticipant!);
+      });
     } else if (ev.case == 'localTrackUnpublished') {
       const publication = this.localParticipant!.trackPublications.get(ev.value.publicationSid!);
       this.localParticipant!.trackPublications.delete(ev.value.publicationSid!);


packages/livekit-rtc/src/room.ts
```diff
@@ -334,8 +334,10 @@ export class Room extends (EventEmitter as new () => TypedEmitter<RoomCallbacks>
       participant!.info.disconnectReason = ev.value.disconnectReason;
       this.emit(RoomEvent.ParticipantDisconnected, participant!);
     } else if (ev.case == 'localTrackPublished') {
-      const publication = this.localParticipant!.trackPublications.get(ev.value.trackSid!);
-      this.emit(RoomEvent.LocalTrackPublished, publication!, this.localParticipant!);
+      setImmediate(() => {
```
Shubhrakanti, line 337: let's add a comment on why we use `setImmediate` here.

Shubhrakanti (commented):
Can you please write what the root cause is, why this is only an issue only with bun and why `setImmediate` solved it?

packages/livekit-rtc/src/room.ts
pzerelles: Sure, the root cause is, that the `localTrackPublished` event is processed before `this.localParticipant.trackPublications` has the track in the map.

This happens, because `publishTrack` in `participant.ts` sends the `publishTrack` request to the FfiClient. Then it waits for the `publishTrack` event from the FfiClient with the result. The result is used to set the track in the `trackPublications` map.

Right after the `publishTrack`event is emitted by the FfiClient, the `localTrackPublished` event is also emitted. Since the `publishTrack` function in `participant.ts` is async and awaits the arrival of the `publishTrack` event, it is not guaranteed that this code will resume before the event loop has finished with other things, like calling the `localTrackPublished` handler. 

By using setImmediate() in the `localTrackPublished` handler, we ensure that the event loop has processed the awaited promises before we try to access the `trackPublications` map. 

Since I don't see the whole picture, this might be solved by not relying on execution order in an asynchronous scenario like it is the case, and refactoring might be a better solution. But this is maybe a bigger thing and using `setImmediate` is a quick fix for the current version.

This is a problem with Bun currently, because it seems to process the event loop differently. But since there is no guarantee, this could break in a future version of NodeJS, too.

lukasIO: > using setImmediate is a quick fix for the current version

this is the only thing I'm a bit concerned about. It doesn't really address the underlying issue that what we'd want to be doing is consecutively process each incoming event so that execution order is guaranteed. 

By `setImmediate`ing one of them, we just potentially kick the can down the road

pzerelles: This could be a general problem the way events are handled with `waitFor`. The event handler should directly perform what is needed, not just resolve a promise that causes another function to continue, because the execution order is not guaranteed for that. One way to handle this maybe is to change `waitFor` in `publishTrack` to an event handler that adds the `track` to `trackPublications`.

tobowers:
would love to get this in, even in a partial fix state because we run into this error *all the time* and would be amazing to get it cleaned up while working out how to properly do events one after the other.

PR #547, opened 2025-10-08
Ensure FFI events are processed sequentially

lukasIO:
For some actions (i.e. publishing + unpublishing a track) the local participant needs to acquire the processing lock to ensure any state updates are done before emitting other events on the room level.

### DIFF
--- a/packages/livekit-rtc/src/participant.ts
+++ b/packages/livekit-rtc/src/participant.ts
@@ -155,7 +155,14 @@ export type DataPublishOptions = {
 export class LocalParticipant extends Participant {
   private rpcHandlers: Map<string, (data: RpcInvocationData) => Promise<string>> = new Map();
 
+  private ffiEventLock: Mutex;
+
   trackPublications: Map<string, LocalTrackPublication> = new Map();
+
+  constructor(info: OwnedParticipant, ffiEventLock: Mutex) {
+    super(info);
+    this.ffiEventLock = ffiEventLock;
+  }
 
   async publishData(data: Uint8Array, options: DataPublishOptions) {
     const req = new PublishDataRequest({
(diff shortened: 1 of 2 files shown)


packages/livekit-rtc/src/participant.ts
```diff
@@ -655,51 +662,62 @@ export class LocalParticipant extends Participant {
...
+
     const res = FfiClient.instance.request<PublishTrackResponse>({
       message: { case: 'publishTrack', value: req },
     });
 
-    const cb = await FfiClient.instance.waitFor<PublishTrackCallback>((ev) => {
-      return ev.message.case == 'publishTrack' && ev.message.value.asyncId == res.asyncId;
-    });
+    try {
+      const cb = await FfiClient.instance.waitFor<PublishTrackCallback>((ev) => {
```
xianshijing-lk, line 672: [thought(non-blocking)](https://pullpo.io/cc?l=thought&d=non-blocking) 
curiously, does FfiClient.instance.request guarantee that it will trigger a PublishTrackCallback later ?


There is no need to change it now, but just an idea, I wonder if it makes sense to do things like
1. prepare a waiter
2. roomEventLock.lock()
3. FfiClient.instance.request
4. waiter.setAsyncId(res.asyncId)
5.  waiter will wait with some timeout, like await waiter.wait({ timeoutMs: 50 }); // not sure if that will bring more flakiness or less here.

packages/livekit-rtc/src/room.ts
```diff
@@ -279,242 +289,253 @@ export class Room extends (EventEmitter as new () => TypedEmitter<RoomCallbacks>
...
-      } else if (trackInfo.kind == TrackKind.KIND_AUDIO) {
-        publication.track = new RemoteAudioTrack(ownedTrack);
+      const ev = ffiEvent.message.value.message;
+      if (process.env.LIVEKIT_DEBUG_LOG_ROOM_EVENTS) {
+        console.log('Room event:', ev);
       }
+      if (ev.case == 'participantConnected') {
+        const participant = this.createRemoteParticipant(ev.value.info!);
+        this.remoteParticipants.set(participant.identity!, participant);
+        this.emit(RoomEvent.ParticipantConnected, participant);
```
xianshijing-lk, line 343: [suggestion(non-blocking)](https://pullpo.io/cc?l=suggestion&d=non-blocking) 
I have some concerns on this emit event since it is fired under the lock, and I wonder if it is a better pattern if we do things like 

const unlock = await this.roomEventLock.lock();
const toEmit: Array<() => void> = [];

try {
  // Push all the emit events in sequence to the array
  toEmit.push(...);
} finally {
  unlock();
}

// Fire the emit events
for (const fire of toEmit) {
  try {
    fire(); // consider await fire() ? if we need them to be synchronous
  } catch ...
}


it is a suggestion, no need to work on it if you think the emit events are safe anyway.

lukasIO: got it, I'd be interested to know what your concern is with the event firing under the lock?

That would help me decide whether or not to leave it as is or implement your suggestion

packages/livekit-rtc/src/participant.ts
lukasIO: > curiously, does FfiClient.instance.request guarantee that it will trigger a PublishTrackCallback later ?

I think so, yes. Not sure what the actual guarantees are though or rather how they are enforced.

PR #549, opened 2025-10-09
Add support for data channel encryption

lukasIO:
also exposes participantEncryptionChanged event and updates rust sdks (currently not to a new FFI build, which needs to get fixed)

### DIFF
--- a/.changeset/witty-olives-explain.md
+++ b/.changeset/witty-olives-explain.md
@@ -1,0 +1,5 @@
+---
+"@livekit/rtc-node": patch
+---
+
+Add support for data channel encryption


packages/livekit-rtc/src/room.ts
```diff
@@ -174,7 +179,10 @@ export class Room extends (EventEmitter as new () => TypedEmitter<RoomCallbacks>
    */
   async connect(url: string, token: string, opts?: RoomOptions) {
     const options = { ...defaultRoomOptions, ...opts };
-    const e2eeOptions = { ...defaultE2EEOptions, ...options.e2ee };
+    const e2eeEnabled = options.encryption || options.e2ee;
+    const e2eeOptions = options.encryption
+      ? { ...defaultE2EEOptions, ...options.encryption }
+      : { ...defaultE2EEOptions, ...options.e2ee };
```
xianshijing-lk, line 192: will it be safe if options.e2ee is undefined here ? 

Specifically, when both options.encryption and options.e2ee are undefined

lukasIO: you mean because of the spread syntax? 
That's not a problem.

PR #560, opened 2025-10-20
Add optional timeouts to clients in server APIs

### LINKED ISSUE #548: Expose timeout option on Server SDK clients?
### Describe the problem

In the server SDKs, I would like to be able to provide a `timeout` to the client request methods so that I can easily abort requests (and also have some idea of the time bound on them).

### Describe the proposed solution

I'm specifically looking at RoomServiceClient. I see the underlying RPC client request() accepts a timeout, but as far as I see, it's never supplied ~and doesn't have a default value~.

I don't need fine-grained control; a "timeout" option on the Client constructor which applied to all its requests would probably suffice.

### Alternatives considered

Does LiveKit Cloud document request timeouts from the server?

### Importance

nice to have

### Additional Information

_No response_

bryfox:
This adds a timeout parameter as part of a new `options` argument to each Server API client.

The RPC client constructor is modified to accept this timeout option, and applies it to all requests if supplied. This interface changes in a breaking way, but `TwirpRPC` isn't exported from the module.

The timeout value is expected to be seconds, which matches the existing timeouts in the RoomServiceClient. There's no up-front validation of the input values; an invalid value will result in an error thrown by node.js when used.

Resolves #548.

### DIFF
--- a/.changeset/honest-apes-sparkle.md
+++ b/.changeset/honest-apes-sparkle.md
@@ -1,0 +1,5 @@
+---
+"livekit-server-sdk": patch
+---
+
+Add optional timeouts to clients in server APIs


packages/livekit-server-sdk/src/AgentDispatchClient.ts
```diff
@@ -29,10 +33,15 @@ export class AgentDispatchClient extends ServiceBase {
    * @param host - hostname including protocol. i.e. 'https://<project>.livekit.cloud'
    * @param apiKey - API Key, can be set in env var LIVEKIT_API_KEY
    * @param secret - API Secret, can be set in env var LIVEKIT_API_SECRET
+   * @param options - client options
+   * @param options.requestTimeout - optional timeout, in seconds, for all server requests
```
lukasIO: Instead of describing the usage here, could you do it just above the field definition in the `ClientOptions` type? 

Would also be nice if we could share the `ClientOptions` type across different implementations instead of redefining it in each client module.

packages/livekit-server-sdk/src/TwirpRPC.ts
```diff
@@ -5,7 +5,15 @@ import type { JsonValue } from '@bufbuild/protobuf';
 
 // twirp RPC adapter for client implementation
 
+type Options = {
+  /** Prefix for the RPC requests */
+  prefix?: string;
+  /** Timeout for fetch requests, in seconds. Must resolve to a positive integer millisecond value. */
```
lukasIO: The "must resolve to a positive integer millisecond value" part feels slightly confusing to me.

lukasIO (commented):
Thanks for the PR, just two minor things

packages/livekit-server-sdk/src/TwirpRPC.ts
bryfox: I re-worded this to be less specific and hopefully less confusing.

PR #567, opened 2025-11-13
AsyncQueue-based AudioMixer

### LINKED ISSUE #566: AudioMixer is missing (but implementation exists :))
### Select which package(s) are affected

\@livekit/rtc-node

### Describe the bug

I found https://github.com/livekit/node-sdks/tree/brian/audio-mixer with a working implementaiton already,
what's blocking it from bringin it to main? :) 

Thanks
Simon

### Reproduction

-

### Logs

```shell

```

### System Info

```shell
-
```

### LiveKit server version

LiveKit cloud

### Severity

serious, but I can work around it

### Additional Information

_No response_

simllll:
based on the work of @toubatbrian

manually tested + added test suite

VERY cpu friendly now too ;-)

fixes #566

### DIFF
--- a/packages/livekit-rtc/src/audio_mixer.ts
+++ b/packages/livekit-rtc/src/audio_mixer.ts
@@ -1,8 +1,8 @@
 // SPDX-FileCopyrightText: 2025 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-import { AudioFrame } from './audio_frame.js';
 import { AsyncQueue } from './async_queue.js';
+import { AudioFrame } from './audio_frame.js';
 
 // Re-export AsyncQueue for backward compatibility
 export { AsyncQueue } from './async_queue.js';
--- a/packages/livekit-rtc/src/async_queue.ts
+++ b/packages/livekit-rtc/src/async_queue.ts
@@ -65,7 +65,9 @@ export class AsyncQueue<T> {
   close() {
     this.closed = true;
     // Reject all waiting producers with an error
-    this.waitingProducers.toArray().forEach((producer) => producer.reject(new Error('Queue closed')));
+    this.waitingProducers
+      .toArray()
+      .forEach((producer) => producer.reject(new Error('Queue closed')));
     // Resolve all waiting consumers so they can see the queue is closed
     this.waitingConsumers.toArray().forEach((resolve) => resolve());
     this.waitingProducers.clear();


packages/livekit-rtc/src/audio_mixer.ts
```diff
@@ -0,0 +1,501 @@
...
+  }
+
+  get(): T | undefined {
+    const item = this.items.shift();
+    if (this.waitingProducers.length > 0) {
+      const resolve = this.waitingProducers.shift()!;
+      resolve(); // wakes up one waiting producer
+    }
+    return item;
+  }
```
toubatbrian: Should we use a more efficient data structure, like an actual Queue?

packages/livekit-rtc/src/audio_mixer.ts
```diff
@@ -0,0 +1,501 @@
...
+        console.log('AudioMixer: No streams available, waiting...');
+        // Wait for a stream to be added (signal queue will have an item)
+        await this.streamSignal.waitForItem();
+        // Consume the signal
+        this.streamSignal.get();
+        console.log('AudioMixer: Woken up, checking streams again...');
+        continue;
+      }
+
+      console.log(`AudioMixer: Processing ${this.streams.size} streams...`);
```
toubatbrian: Once we finish testing, let's remove the `console.log`

packages/livekit-rtc/src/audio_mixer.ts
simllll: the reason why I introduced AsyncQueue is that this queue has consomers und producers to queue. So we can remove a lot of logic from the rest of the code, by just wating (e.g. await put(..)) till the queue is not full anymore, or the queue is not empty again.

packages/livekit-rtc/src/audio_mixer.ts
simllll: Oh I missed that, sorry

packages/livekit-rtc/src/audio_mixer.ts
simllll: added a comment in the code that explains it too.

packages/livekit-rtc/src/audio_mixer.ts
simllll: removed the console logs

packages/livekit-rtc/src/audio_mixer.ts
toubatbrian: oh just to clarify I mean specifically the `this.waitingConsumers.shift()` operation which is O(n), can we use something like a deque for the consumer and producers: https://datastructures-js.info/docs/deque? I think overall the idea of AsyncQueue looks great!

simllll: Ah now I understand :)

Not sure if it's worth it to add a new dependency? I could also just implement a simple O(1) list-based queue internally? I'm following your suggestion, let me know ;-)

toubatbrian: Yeah for me a list-based queue is also fine. @lukasIO let me know your thought

simllll: ` import { Deque } from '@datastructures-js/deque'` it is ;-) less code to mantain is always better. and if a new dependency is okay, then let's do it. it's quite small too.

packages/livekit-rtc/src/audio_mixer.ts
```diff
@@ -0,0 +1,472 @@
...
+      resolve();
+    }
+  }
+
+  get(): T | undefined {
+    const item = this.items.shift();
+    if (this.waitingProducers.size() > 0) {
+      const resolve = this.waitingProducers.popFront()!;
+      resolve(); // wakes up one waiting producer
+    }
```
toubatbrian: can we add some tests to AsyncQueue as well?

packages/livekit-rtc/package.json
```diff
@@ -47,18 +47,19 @@
   },
   "dependencies": {
     "@bufbuild/protobuf": "^1.10.1",
+    "@datastructures-js/deque": "^1.0.8",
```
ai-simple-ai: This package has only one maintainer and only about 2 thousand downloads per week. Does it make sense to vendor this dependency into the repository to make sure it doesn't become an attack vector?

At the very least, does it make sense to pin the specific version of the repository without the semver?

lukasIO: +1 for pinning

toubatbrian: > This package has only one maintainer and only about 2 thousand downloads per week. Does it make sense to vendor this dependency into the repository to make sure it doesn't become an attack vector?

Good point. Given that we only use the `Deque` data structure from their package. Perhaps it make sense if we directly copy the code source file? https://github.com/datastructures-js/deque/blob/main/src/deque.js and add a reference comment https://github.com/datastructures-js/deque/blob/main/LICENSE

PR #573, opened 2025-11-18
feat(connector): initial client impl

anunaym14:
Requires: https://github.com/livekit/protocol/pull/1311

### DIFF
--- a/packages/livekit-server-sdk/src/ConnectorClient.ts
+++ b/packages/livekit-server-sdk/src/ConnectorClient.ts
@@ -1,12 +1,15 @@
 // SPDX-FileCopyrightText: 2025 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-import type { RoomAgentDispatch, SessionDescription } from '@livekit/protocol';
+import type {
+  ConnectTwilioCallRequest_TwilioCallDirection,
+  RoomAgentDispatch,
+  SessionDescription,
+} from '@livekit/protocol';
 import {
   AcceptWhatsAppCallRequest,
   AcceptWhatsAppCallResponse,
   ConnectTwilioCallRequest,
-  ConnectTwilioCallRequest_TwilioCallDirection,
   ConnectTwilioCallResponse,
   ConnectWhatsAppCallRequest,
   ConnectWhatsAppCallResponse,


packages/livekit-server-sdk/src/ConnectorClient.ts
```diff
@@ -0,0 +1,283 @@
...
+  /** Optional - User-defined attributes. Will be attached to a created Participant in the room. */
+  participantAttributes?: { [key: string]: string };
+  /** Optional - Country where the call terminates as ISO 3166-1 alpha-2 */
+  destinationCountry?: string;
+}
+
+// Twilio types
+export interface ConnectTwilioCallOptions {
+  /** The direction of the call */
+  twilioCallDirection: ConnectTwilioCallRequest_TwilioCallDirection;
```
davidzhao, line 87: we need to re-export this in `index.ts` to ensure users could do `import type { ConnectTwilioCallRequest_TwilioCallDirection } from "livekit-server-sdk"`

otherwise they'd have to do the awkward `import ... from @livekit/protocol`

packages/livekit-server-sdk/src/ConnectorClient.ts
```diff
@@ -0,0 +1,283 @@
...
+    this.rpc = new TwirpRpc(host, livekitPackage, rpcOptions);
+  }
+
+  /**
+   * Initiate an outbound WhatsApp call
+   *
+   * @param options - WhatsApp call options
+   * @returns Promise containing the WhatsApp call ID and room name
+   */
+  async dialWhatsAppCall(options: DialWhatsAppCallOptions): Promise<DialWhatsAppCallResponse> {
```
davidzhao, line 130: `DialWhatsAppCallResponse` and the other response types need to be re-exported as well

packages/livekit-server-sdk/src/ConnectorClient.ts
anunaym14: done!

packages/livekit-server-sdk/src/ConnectorClient.ts
anunaym14: done!

packages/livekit-server-sdk/src/ConnectorClient.ts
```diff
@@ -0,0 +1,279 @@
+// SPDX-FileCopyrightText: 2025 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import type { RoomAgentDispatch, SessionDescription } from '@livekit/protocol';
+import {
```
davidzhao, line 9: please correct for this comment. some of them needs to be type imports

anunaym14: which comment are you talking about?

type imports should be fixed now!

davidzhao (approved):
lg!

PR #579, opened 2025-12-04
Hold lock on all ffi events

lukasIO:
for reviewing, enable "hide whitespace" in the review tab. 

changes are minimal, but due to indentation levels changing the diff appears quite large.

### DIFF
--- a/.changeset/cool-peaches-flash.md
+++ b/.changeset/cool-peaches-flash.md
@@ -1,0 +1,5 @@
+---
+"@livekit/rtc-node": patch
+---
+
+Hold lock on all ffi events


packages/livekit-rtc/src/room.ts
```diff
@@ -302,321 +302,319 @@ export class Room extends (EventEmitter as new () => TypedEmitter<RoomCallbacks>
...
+      const participant = this.requireRemoteParticipant(ev.value.participantIdentity!);
+      const publication = participant.trackPublications.get(ev.value.publicationSid!);
+      participant.trackPublications.delete(ev.value.publicationSid!);
+      if (publication) {
+        this.emit(RoomEvent.TrackUnpublished, publication, participant);
+      } else {
+        console.warn(`RoomEvent.TrackUnpublished: Could not find publication`);
+      }
+    } else if (ev.case == 'trackSubscribed') {
+      console.info('ffi track subscribed event received');
```
1egoman: nitpick: Was this log intended to be left here or was this only for debugging?

lukasIO: forgot to push up the commit that removed this, thanks for catching that!

PR #580, opened 2025-12-10
Use trusted publishing in release CI

lukasIO:


.github/workflows/ci.yml
```diff
@@ -244,7 +250,6 @@ jobs:
           publish: pnpm ci:publish
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
toubatbrian, line 247: So this means we no longer need any secrets for trusted publish?

toubatbrian (approved):

.github/workflows/ci.yml
lukasIO: exactly

.github/workflows/ci.yml
```diff
@@ -72,7 +78,7 @@ jobs:
     name: Test
     strategy:
       matrix:
-        node-version: [18, 20, 22, latest]
```
1egoman, line 75: @lukasIO

suggestion: The [lowest supported node version of livekit-rtc package is v18](https://github.com/livekit/node-sdks/blob/081df95783da838ca998e8580c878e486359e540/packages/livekit-rtc/package.json#L72) so it would probably be worthwhile to keep the tests running on that version. Or alternatively, maybe it's worth updating everything to node 20 if it's ok to drop node 18 support at this point.

lukasIO: security support for node 18 ended 6 months ago, so probably worth updating everything to 20

PR #583, opened 2025-12-10
Add frame processor support for audio streams (node version)

1egoman:
This is a copy of https://github.com/livekit/python-sdks/pull/533, adapted for node.

I'm letting @lukasIO take the lead on this, and will adjust this once the python version gets merged to match the same interface.

### DIFF
--- a/packages/livekit-rtc/src/frame_processor.ts
+++ b/packages/livekit-rtc/src/frame_processor.ts
@@ -1,8 +1,8 @@
 // SPDX-FileCopyrightText: 2025 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-import { AudioFrame } from './audio_frame.js';
-import { VideoFrame } from './video_frame.js';
+import { type AudioFrame } from './audio_frame.js';
+import { type VideoFrame } from './video_frame.js';
 
 export type FrameProcessorStreamInfo = {
   roomName: string;
@@ -19,7 +19,9 @@ export abstract class FrameProcessor<Frame extends VideoFrame | AudioFrame> {
   abstract isEnabled(): boolean;
   abstract setEnabled(enabled: boolean): void;
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   onStreamInfoUpdated(_info: FrameProcessorStreamInfo): void {}
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   onCredentialsUpdated(_credentials: FrameProcessorCredentials): void {}
 
   abstract process(frame: Frame): Frame;
--- a/packages/livekit-rtc/src/audio_stream.ts
+++ b/packages/livekit-rtc/src/audio_stream.ts
@@ -20,6 +20,7 @@ export interface AudioStreamOptions {
 
 export interface NoiseCancellationOptions {
   moduleId: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   options: Record<string, any>;
 }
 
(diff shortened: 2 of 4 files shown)


examples/receive-audio-through-aic-filter/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "type": "module",
+  "main": "index.ts",
+  "scripts": {
+    "build": "tsc --incremental",
+    "lint": "eslint -f unix \"**/*.ts\""
+  },
+  "keywords": [],
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@livekit/audio-filter-uniffi": "link:/Users/ryan/w/livekit/aic-noise-filter/target/packages/node",
```
1egoman: todo: This needs to be updated once the aic audio filter package gets published, and is what seems to be keeping teh ci build from passing at the moment.

packages/livekit-rtc/src/audio_frame.ts
```diff
@@ -11,6 +11,8 @@ export class AudioFrame {
   channels: number;
   samplesPerChannel: number;
 
+  private _userdata: Record<string, any> = {};
```
lukasIO: nit: `unknown` instead of `any`

packages/livekit-rtc/src/frame_processor.ts
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export abstract class FrameProcessor<Frame extends VideoFrame | AudioFrame> {
+  abstract isEnabled(): boolean;
+  abstract setEnabled(enabled: boolean): void;
+
+  abstract onStreamInfoUpdated(info: FrameProcessorStreamInfo): void;
+  abstract onCredentialsUpdated(credentials: FrameProcessorCredentials): void;
+
+  abstract process(frame: Frame): Frame;
```
lukasIO, line 27: question: in node (in contrast to python) optional async handling would be a lot more straight forward. Should we make this async?

packages/livekit-rtc/src/frame_processor.ts
```diff
@@ -0,0 +1,27 @@
...
+  token: string;
+  url: string;
+};
+
+export abstract class FrameProcessor<Frame extends VideoFrame | AudioFrame> {
+  abstract isEnabled(): boolean;
+  abstract setEnabled(enabled: boolean): void;
+
+  abstract onStreamInfoUpdated(info: FrameProcessorStreamInfo): void;
+  abstract onCredentialsUpdated(credentials: FrameProcessorCredentials): void;
```
lukasIO: issue: these two have a default implementation in python and aren't treated as abstract. We should make them non-required also for node.

packages/livekit-rtc/src/frame_processor.ts
1egoman: Unless you have a strong opinion, I think I'd prefer to mirror the python interface as much as possible. Making this async would be pretty easy to do at a later date if we end up needing it for some reason.

I'm not sure which function you are pointing to directly (the section of highlighted code shows `onStreamInfoUpdated`, `onCredentialsUpdated`, and `process`), but I also wanted to mention that I think making `process` async probably isn't a good idea, since it needs to run for every audio frame, so if it was async, it could be very easy to use it in such a way where it doesn't resolve in time for the next audio frame, leading to overlapping frames being processed at the same time and potentially some confusing edge cases / performance degredation.

packages/livekit-rtc/src/frame_processor.ts
1egoman: Ah, I see, I think I missed that. Fixed!

packages/livekit-rtc/src/frame_processor.ts
lukasIO: > Making this async would be pretty easy to do at a later date

but that would be a breaking change, I think. 


> I'm not sure which function you are pointing to directly

was only referring to the process function

PR #584, opened 2025-12-18
Set token NotBefore Time to now

cnderrauber:
cloud will check NotBefore timestamp to revoke tokens


packages/livekit-server-sdk/src/AccessToken.ts
```diff
@@ -203,7 +203,7 @@ export class AccessToken {
       .setProtectedHeader({ alg: 'HS256' })
       .setIssuer(this.apiKey)
       .setExpirationTime(this.ttl)
-      .setNotBefore(0);
+      .setNotBefore(new Date());
```
lukasIO, line 206: looks good! 

was also curious why your previous attempt failed in the test and it turns out jose expects the unix timestamp in seconds and `Date.now()` returns millis

PR #591, opened 2026-01-03
add e2e tests

davidzhao:


packages/livekit-rtc/src/tests/e2e.test.ts
```diff
@@ -0,0 +1,514 @@
...
+
+  return bestLag > 0 ? sampleRate / bestLag : 0;
+}
+
+describeE2E('livekit-rtc e2e', () => {
+  afterAll(async () => {
+    await dispose();
+  });
+
+  it(
```
lukasIO, line 205: By default vite only runs test _files_ in parallel, if we want to speed the runs up we could use `it.concurrent` instead in order to run tests in parallel, too. 

this needs `config.test.maxConcurrency` set as well in the vite config.

packages/livekit-rtc/src/tests/e2e.test.ts
```diff
@@ -0,0 +1,514 @@
...
+  RoomEvent,
+  RpcError,
+  TrackPublishOptions,
+  TrackSource,
+  dispose,
+} from '../index.js';
+
+const hasE2EEnv =
+  !!process.env.LIVEKIT_URL && !!process.env.LIVEKIT_API_KEY && !!process.env.LIVEKIT_API_SECRET;
+const describeE2E = hasE2EEnv ? describe : describe.skip;
```
lukasIO, line 25: what scenarios did you have in mind for skipping?
Wondering if this might hide issues as the workflow run will look "clean" even though the tests didn't run at all.

packages/livekit-rtc/src/tests/e2e.test.ts
```diff
@@ -0,0 +1,514 @@
...
+          destinationIdentity: 'unknown-participant',
+          method,
+          payload,
+          responseTimeout: 500,
+        }),
+      ).rejects.toMatchObject({ code: RpcError.ErrorCode.CONNECTION_TIMEOUT });
+
+      await Promise.all(rooms.map((r) => r.disconnect()));
+    },
+    testTimeoutMs,
```
lukasIO: running against a local server this test regularly times out for me. 

Most time is spent on the last assertion for the connection timeout. 
What do you think about splitting these up in its own RPC suite with each assertion being its own test?

packages/livekit-rtc/src/tests/e2e.test.ts
davidzhao: it'll skip if the livekit env vars aren't defined or empty.. which would be the case with forks / community contributed PRs.

this matches most of our tests where the full e2e isn't ran on community patches

packages/livekit-rtc/src/tests/e2e.test.ts
davidzhao: interesting.. this seems like a bug right? `performRpc` is supposed to timeout in 500ms..

should `responseTimeout` also be applied to the connection as well?

lukasIO: we hardcode a max_roundtrip_latency and a min_effective_timeout, so we always end up with an 8s minimum, I think

https://github.com/livekit/rust-sdks/blob/8b69e0838dafa3f33171f1f55836ba3ed9c728f6/livekit/src/room/participant/local_participant.rs#L745-L746

davidzhao: ah.. this is not great.. ok. I'll change the timeout for now to unblock the test

PR #647, opened 2026-04-08
Data tracks support

ladvoc:
Resolves BOT-281

### DIFF
--- a/packages/livekit-rtc/src/data_tracks/local.ts
+++ b/packages/livekit-rtc/src/data_tracks/local.ts
@@ -17,6 +17,7 @@ import { FfiClient, FfiHandle } from '../ffi_client.js';
 import type { DataTrackFrame, DataTrackInfo } from './types.js';
 import { DataTrackPushFrameError } from './types.js';
 
+/** Data track published by the local participant. */
 export class LocalDataTrack {
   private _info: DataTrackInfo;
   private ffiHandle: FfiHandle;
@@ -31,10 +32,12 @@ export class LocalDataTrack {
     this.ffiHandle = new FfiHandle(ownedTrack.handle!.id!);
   }
 
+  /** Information about the data track. */
   get info(): DataTrackInfo {
     return this._info;
   }
 
+  /** Whether or not the track is still published. */
   isPublished(): boolean {
     const res = FfiClient.instance.request<LocalDataTrackIsPublishedResponse>({
       message: {
@@ -47,6 +50,18 @@ export class LocalDataTrack {
     return res.isPublished!;
   }
 
+  /**
+   * Try pushing a frame to subscribers of the track.
+   *
+   * See {@link DataTrackFrame} for how to construct a frame and attach metadata.
+   *
+   * Pushing a frame can fail for several reasons:
+   *
+   * - The track has been unpublished by the local participant or SFU
+   * - The room is no longer connected
+   *
+   * @throws {@link DataTrackPushFrameError} If the push fails.
+   */
   tryPush(frame: DataTrackFrame): void {
     const protoFrame = new ProtoDataTrackFrame({
       payload: frame.payload,
@@ -68,6 +83,10 @@ export class LocalDataTrack {
     }
   }
 
+  /**
+   * Unpublish the track from the SFU. Once this is called, any further calls to
+   * {@link tryPush} will fail.
+   */
   async unpublish(): Promise<void> {
     FfiClient.instance.request({
       message: {
--- a/packages/livekit-rtc/src/data_tracks/local.ts
+++ b/packages/livekit-rtc/src/data_tracks/local.ts
@@ -1,0 +1,81 @@
+// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import type {
+  LocalDataTrackIsPublishedResponse,
+  LocalDataTrackTryPushResponse,
+  OwnedLocalDataTrack,
+} from '@livekit/rtc-ffi-bindings';
+import {
+  DataTrackFrame as ProtoDataTrackFrame,
+  LocalDataTrackIsPublishedRequest,
+  LocalDataTrackTryPushRequest,
+  LocalDataTrackUnpublishRequest,
+} from '@livekit/rtc-ffi-bindings';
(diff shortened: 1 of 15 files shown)


packages/livekit-rtc/src/data_tracks/local.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (res.error) {
+      throw new DataTrackPushFrameError(res.error.message!);
+    }
+  }
+
+  /**
+   * Unpublish the track from the SFU. Once this is called, any further calls to
+   * {@link tryPush} will fail.
+   */
+  async unpublish(): Promise<void> {
```
lukasIO, line 89: if this is async, do we need to await a dataTrackUnpublished response ?

packages/livekit-rtc/src/data_tracks/local.ts
```diff
@@ -0,0 +1,99 @@
...
+   * See {@link DataTrackFrame} for how to construct a frame and attach metadata.
+   *
+   * Pushing a frame can fail for several reasons:
+   *
+   * - The track has been unpublished by the local participant or SFU
+   * - The room is no longer connected
+   *
+   * @throws {@link DataTrackPushFrameError} If the push fails.
+   */
+  tryPush(frame: DataTrackFrame): void {
```
lukasIO, line 64: thought(non-blocking): I left a similar comment on the JS implementation, it seems a bit counter-intuitive that a `try*` method can throw. 
Although I don't know if there's a good alternative here. If users aren't expected to gain additional insight from the error's message maybe returning a "success" boolean from the method would be sufficient?

packages/livekit-rtc/src/data_tracks/remote.ts
```diff
@@ -0,0 +1,130 @@
...
+        controller.enqueue({
+          payload: protoFrame.payload!,
+          userTimestamp: protoFrame.userTimestamp,
+        });
+        break;
+      }
+      case 'eos': {
+        this.dispose();
+        if (event.detail.value.error) {
+          controller.error(new Error(event.detail.value.error));
```
lukasIO, line 111: suggestion(non-blocking): maybe nice to have a more specific (tagged) error here

packages/livekit-rtc/src/data_tracks/remote.ts
```diff
@@ -0,0 +1,130 @@
...
+      message: {
+        case: 'subscribeDataTrack',
+        value: new SubscribeDataTrackRequest({
+          trackHandle: this.ffiHandle.handle,
+          options: opts,
+        }),
+      },
+    });
+
+    return new ReadableStream<DataTrackFrame>(new DataTrackStreamSource(res.stream!), {
```
lukasIO, line 66: question: should we be worried about the fact that each subscribe call pushes the same frames across the FFI boundary each time?
Not sure if there's a good way to re-use an existing source. 
We could also think about `tee`ing the stream but that comes with it's own pitfalls when reading at different rates in different places.

packages/livekit-rtc/src/data_tracks/local.ts
ladvoc: Currently, the Rust implementation has a sync unpublish method. I made this async for future proofing like we did in Python.

packages/livekit-rtc/src/data_tracks/remote.ts
ladvoc: That's definitely something we can do here—the Protobuf now has enums that map to the corresponding error cases from Rust. However, I wasn't sure the nicest way to expose this in Node and if we want to adopt the new error handling patterns from the JS SDK.

packages/livekit-rtc/src/data_tracks/remote.ts
ladvoc: Definitely something to consider, and something that also applies to Python, C++, and Unity. If there are multiple subscriptions, you are right that the same frames will need to be serialized/deserialized more than once. I wonder if the performance overhead from this is enough to justify implementing a teeing solution for each FFI client—this would be more performant at the cost of adding additional complexity to each FFI client implementation. Currently, teeing is implemented on the Rust side so each FFI client gets that behavior automatically.

packages/livekit-rtc/src/data_tracks/local.ts
ladvoc: In contrast to the web implementation, the error reasons here might be more useful and worth exposing. Under the hood, this sends the frame to a dedicated task (one for each track) for processing (E2EE, compression in a future release, etc.) and packetization before sending. If the channel used for sending frames to this task fills up (i.e., in the case the user is pushing too fast), they would get an error here.

lukasIO: makes sense, my slight reservation about it being called `try*` in that case remains

packages/livekit-rtc/src/data_tracks/remote.ts
lukasIO: I'm all for more specific (and tagged) errors! If you think it would fall in `suggestion(if-minor)` then I'd say go for it ;)

packages/livekit-rtc/src/data_tracks/remote.ts
lukasIO: > I wonder if the performance overhead from this is enough

yeah, would be nice to get a feel for this with some numbers. Definitely nothing blocking right now, but would be a good follow up to check at what point this would start to become a bottleneck

packages/livekit-rtc/src/participant.ts
```diff
@@ -720,6 +731,39 @@ export class LocalParticipant extends Participant {
     }
   }
 
+  /**
+   * Publishes a data track.
+   *
+   * @returns The published data track. Use {@link LocalDataTrack.tryPush} to send data frames.
+   * @throws {@link PublishDataTrackError} if there is an error publishing the data track.
```
lukasIO, line 738: Let's add some hint in the comment for minimum OSS server version required

ladvoc: Resolved in [ab73fc6](https://github.com/livekit/node-sdks/pull/647/commits/ab73fc61bfb273a761c34c258df3e73c7f74b78d)

PR #649, opened 2026-04-14
fix: prevent listener leak in FfiClient.waitFor

LautaroPetaccio:
The `waitFor` method on `FfiClient` registers an event listener that is only removed when the predicate matches. If the native FFI layer never sends the expected callback — due to network disconnect, native crash, or `room.disconnect()` being called while operations are in-flight — the listener stays attached to the process-global `FfiClient` singleton forever. Each orphaned listener continues running its predicate against every future FFI event, leaking memory and wasting CPU.

This change adds two cleanup mechanisms to `waitFor`:

- **Timeout** (default 10s via `FfiClient.defaultTimeout`): if the expected event never arrives, the listener is removed and the promise rejects with a descriptive error. The default is configurable per-call.
- **AbortSignal**: callers can pass an `AbortSignal` to cancel pending `waitFor` operations externally, enabling patterns like cancelling all in-flight operations on disconnect.

Both parameters are optional and backward-compatible — existing call sites continue to work without changes, now with automatic cleanup after 10 seconds.

### DIFF
--- a/.changeset/waitfor-listener-leak.md
+++ b/.changeset/waitfor-listener-leak.md
@@ -1,0 +1,5 @@
+---
+'@livekit/rtc-node': patch
+---
+
+Add timeout and abort support to FfiClient.waitFor to prevent listener leaks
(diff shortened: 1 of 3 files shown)


LautaroPetaccio:
Closing: equivalent fix already merged upstream in #636

PR #650, opened 2026-04-14
Fix ReadableStream lock leak in stream readers

LautaroPetaccio:
Both `ByteStreamReader` and `TextStreamReader` implement `[Symbol.asyncIterator]` with a `return()` method that calls `reader.releaseLock()`. However, per the ECMAScript spec, `for await...of` only calls `return()` when the loop exits **abruptly** (via break, return, or throw from the loop body). When `next()` returns `{ done: true }` — either on normal stream completion or in the error catch block — the loop ends normally and `return()` is never invoked. This leaves the `ReadableStream` permanently locked.

The fix calls `reader.releaseLock()` in `next()` whenever returning `{ done: true }`, covering both the normal completion path and the error path. The `return()` method is wrapped in a try/catch so it remains idempotent when the lock was already released by `next()`.

### DIFF
--- a/.changeset/stream-reader-lock-leak.md
+++ b/.changeset/stream-reader-lock-leak.md
@@ -1,0 +1,5 @@
+---
+'@livekit/rtc-node': patch
+---
+
+Fix ReadableStream lock leak in ByteStreamReader and TextStreamReader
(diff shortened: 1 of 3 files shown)


LautaroPetaccio:
Closing: equivalent fix already merged upstream in #640

PR #651, opened 2026-04-14
Fix stream controller map leak on disconnect

LautaroPetaccio:
## Summary

When a sender's data stream is aborted (e.g., due to a transport failure during `sendChunk`), no `DataStream_Trailer` is sent to the remote side. On the receiver, `handleStreamTrailer()` is the only code path that calls `controller.close()` and removes entries from the `byteStreamControllers` / `textStreamControllers` Maps. Since `room.disconnect()` does not clear these Maps either, the `StreamController` objects (containing header, `ReadableStreamDefaultController`, and `startTime`) leak for the entire lifetime of the Room object.

## Changes

**Receiver side (`room.ts`):**
- Added `cleanupStreamControllers()` that iterates both Maps, errors any open controllers (so consumers get a proper error instead of hanging indefinitely), and clears the entries.
- Called from `disconnect()` before removing event listeners.

**Sender side (`participant.ts`):**
- Changed the `abort()` handlers in both `streamText()` and `streamBytes()` to send a `DataStream_Trailer` with the error reason on a best-effort basis. If the transport is already down, the send fails silently — the receiver-side cleanup handles that case.

### DIFF
--- a/.changeset/stream-controller-map-leak.md
+++ b/.changeset/stream-controller-map-leak.md
@@ -1,0 +1,5 @@
+---
+'@livekit/rtc-node': patch
+---
+
+Fix stream controller map leak on disconnect and abort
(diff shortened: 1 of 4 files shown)


LautaroPetaccio:
Closing: equivalent fix already merged upstream in #633

PR #652, opened 2026-04-14
fix(leak): Memory leak when Room.connect() fails

LautaroPetaccio:
## Summary

When `Room.connect()` fails, the `onFfiEvent` listener registered on the global `FfiClient` singleton is never removed. This causes two problems:

- **The Room instance cannot be garbage collected** because `FfiClient` holds a reference to `this.onFfiEvent`, which closes over the Room.
- **`preConnectEvents` grows unboundedly** because `onFfiEvent` keeps pushing every `FfiEvent` into the array — the guard `!this.localParticipant || !this.ffiHandle || !this.info` is always true for a Room that never finished connecting.
- **Retrying `connect()` on the same Room multiplies the leak**, since each failed attempt adds another copy of the listener.

The fix wraps the connect logic in a `try/catch` so that on any error (whether from `waitFor` rejection or the `ConnectError` thrown on the `'error'` case), we:

1. Remove the `onFfiEvent` listener from `FfiClient`
2. Clear the `preConnectEvents` array

The listener is intentionally **not** removed on the success path — it must stay attached for the room to function normally.

### DIFF
--- a/.changeset/connect-error-listener-leak.md
+++ b/.changeset/connect-error-listener-leak.md
@@ -1,0 +1,5 @@
+---
+'@livekit/rtc-node': patch
+---
+
+Remove FfiClient listener on failed connect to prevent leak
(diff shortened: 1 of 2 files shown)


lukasIO (approved):
this makes sense, thank you for the fix!

PR #655, opened 2026-04-28
fix(rtc): consistent connection state handling

lukasIO:
the connect callback flipped `this.connectionState` without explicitly emitting the event. 

Moving everything to a update handler that emits the event in case of a change.


1egoman (approved):
Makes sense to me, then that event acts as a source of truth!

PR #657, opened 2026-04-29
feat: update agent dispatch and connector client

anunaym14:
also expose kind detail on participant


lukasIO (approved):
lgtm!

Could you add a changeset as well?

PR #666, opened 2026-05-20
ReadyForRoomEvent + update to livekit ffi 0.12.60

stephen-derosa:
# Overview
Send ReadyForRoomEventRequest once async connection is returned from rust.

### Implications
Without this [as of this commit in the rust SDK](https://github.com/livekit/rust-sdks/commit/d74892de57724bd08dc2307b68e4251b8eae2f33) after fulfilling the async connection response, rust will timeout with a panic after 1s of not getting this request


stephen-derosa:
looks like my prettier stage is failing, i ran it locally but i see its unchanged -- maybe my local prettier format (default) is not what the repo uses?
```
❯ pnpm exec prettier --check "packages/livekit-rtc/src/room.ts"
Checking formatting...
All matched files use Prettier code style!
```

lukasIO (approved):
lgtm! 

prettier should pick up the config from this repo, I pushed a fix (was just a import ordering issue), for next time you can try running `pnpm run format:check` and/or `pnpm run format:write` locally and it should work

ladvoc (approved):
LGTM ✅, thanks for adding this @stephen-derosa!

1egoman (approved):
I tested this locally with agents-js (which exercised both `readyForRoomEvent` and RPC v2) and everything worked as expected. In addition, the new "long rpc" case in the rpc example I added also passes.

I think this is ready to go, I will aim to get this released on monday!

PR #668, opened 2026-05-27
Add token revocation timestamp to RemoveParticipant

cnderrauber:
User can provide timestamp to invalidate token whose nbf is before this value.

### DIFF
--- a/.changeset/fuzzy-trains-lose.md
+++ b/.changeset/fuzzy-trains-lose.md
@@ -1,0 +1,5 @@
+---
+'livekit-server-sdk': patch
+---
+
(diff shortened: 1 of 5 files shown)


lukasIO (approved):
lgtm. 

just a clarifying question: if `revokeTokenTs` is not set, will it default to the current time for revocation server side or will revocation only happen if a timestamp is set?

cnderrauber:
> lgtm.
> 
> just a clarifying question: if `revokeTokenTs` is not set, will it default to the current time for revocation server side or will revocation only happen if a timestamp is set?

It will keep the current behavior: the server use `now` with 1min leeway as the timestamp

PR #671, opened 2026-05-27
Add initial support for frame processor usage directly on tracks

1egoman:
Updates the node sdk so that `FrameProcessor`-based noise cancellation providers can be used directly on `AudioStream`, without having to go through the agent's RoomIO to be able to initialize itself with credentials.

For example, with this change, something like the below becomes possible:
```ts
const stream = new AudioStream(track, {
  sampleRate: SAMPLE_RATE,
  numChannels: CHANNELS,
  noiseCancellation: aicAudioEnhancement({ model: 'quailVfL' }),
});
```

The way this works - `Track`s now keep track of which room they are part of (holding a `WeakRef` value). When the room a track is in changes, it computes new frame processor options and sends these to any `AudioStream`s which are associated with the track.

The `noiseCancellationLeaveOpen` parameter allows the agents sdk to construct an `AudioStream` with a frame processor which remains open across the whole session, and won't be auto-closed when the track is closed.

## Todo
- [ ] Test throughly (just ran a happy path test so far)

### DIFF
--- a/packages/livekit-rtc/src/audio_stream.ts
+++ b/packages/livekit-rtc/src/audio_stream.ts
@@ -31,6 +31,7 @@ export interface NoiseCancellationOptions {
   options: Record<string, any>;
 }
 
+/** @internal */
 export class AudioStreamSource implements UnderlyingSource<AudioFrame> {
   private controller?: ReadableStreamDefaultController<AudioFrame>;
   private ffiHandle: FfiHandle;
--- a/packages/livekit-rtc/src/audio_stream.ts
+++ b/packages/livekit-rtc/src/audio_stream.ts
@@ -13,6 +13,12 @@ import type { Track } from './track.js';
 
 export interface AudioStreamOptions {
   noiseCancellation?: NoiseCancellationOptions | FrameProcessor<AudioFrame>;
+  /**
+   * If true and `noiseCancellation` is a {@link FrameProcessor}, leaves the
+   * processor open when the stream closes so the same processor can be reused
+   * with another {@link AudioStream}. Defaults to `false`.
+   */
+  noiseCancellationLeaveOpen?: boolean;
   sampleRate?: number;
   numChannels?: number;
   frameSizeMs?: number;
@@ -31,19 +37,23 @@ class AudioStreamSource implements UnderlyingSource<AudioFrame> {
   private sampleRate: number;
   private numChannels: number;
   private legacyNcOptions?: NoiseCancellationOptions;
-  private frameProcessor?: FrameProcessor<AudioFrame>;
+  private frameProcessor: FrameProcessor<AudioFrame> | null = null;
+  private leaveProcessorOpen = false;
   private frameSizeMs?: number;
+  private track: Track;
 
   constructor(
     track: Track,
     sampleRateOrOptions?: number | AudioStreamOptions,
     numChannels?: number,
   ) {
+    this.track = track;
     if (sampleRateOrOptions !== undefined && typeof sampleRateOrOptions !== 'number') {
       this.sampleRate = sampleRateOrOptions.sampleRate ?? 48000;
       this.numChannels = sampleRateOrOptions.numChannels ?? 1;
       if (isFrameProcessor(sampleRateOrOptions.noiseCancellation)) {
         this.frameProcessor = sampleRateOrOptions.noiseCancellation;
+        this.leaveProcessorOpen = sampleRateOrOptions.noiseCancellationLeaveOpen ?? false;
       } else {
         this.legacyNcOptions = sampleRateOrOptions.noiseCancellation;
       }
@@ -77,6 +87,12 @@ class AudioStreamSource implements UnderlyingSource<AudioFrame> {
     this.ffiHandle = new FfiHandle(res.stream!.handle!.id!);
 
     FfiClient.instance.on(FfiClientEvent.FfiEvent, this.onEvent);
+    track.registerAudioStream(this);
+  }
+
+  /** @internal */
+  get processor(): FrameProcessor<AudioFrame> | null {
+    return this.frameProcessor;
   }
 
   private onEvent = (ev: FfiEvent) => {
@@ -113,8 +129,11 @@ class AudioStreamSource implements UnderlyingSource<AudioFrame> {
         // while buffered frames are still in the ReadableStream queue.
         if (!this.disposed) {
           this.disposed = true;
+          this.track.unregisterAudioStream(this);
           this.ffiHandle.dispose();
-          this.frameProcessor?.close();
+          if (this.frameProcessor && !this.leaveProcessorOpen) {
+            this.frameProcessor.close();
+          }
         }
         break;
     }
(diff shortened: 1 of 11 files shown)


packages/livekit-rtc/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "outDir": "dist",
-    "declarationDir": "dist"
+    "declarationDir": "dist",
+    "lib": ["es2015", "es2021.weakref"]
```
1egoman, line 6: Is this an ok addition to be making here? Or does there need to be a `WeakRef` polyfill which falls back to some sort of a no-op class wrapper?

(in case it's useful, `WeakRef` was introduced in node v16.4: [mdn link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef#browser_compatibility))

1egoman:
Going to wait for some of the conversation on the [python version of this](https://github.com/livekit/python-sdks/pull/679) to shake out before continuing to push this forward.

packages/livekit-rtc/src/track.ts
```diff
@@ -21,9 +23,140 @@ export abstract class Track {
...
+      }
+      if (room) {
+        room.on('tokenRefreshed', this.onRoomTokenRefreshed);
+      }
+    }
+    this.roomRef = room ? new WeakRef(room) : null;
+    for (const stream of this.iterateStreams()) {
+      this.pushProcessorMetadataToStream(stream, room);
+    }
+  }
```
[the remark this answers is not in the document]
1egoman: Should be fixed in [c1e6f43](https://github.com/livekit/node-sdks/pull/671/commits/c1e6f43501fcdc03488c8e748dbc15bed9a51b52)

packages/livekit-rtc/src/track.ts
```diff
@@ -21,9 +23,146 @@ export abstract class Track {
...
+        if (local) {
+          for (const publication of local.trackPublications.values()) {
+            if (publication.sid === trackSid) {
+              identity = local.identity;
+              publicationSid = publication.sid ?? '';
+              break;
+            }
+          }
+        }
+      }
```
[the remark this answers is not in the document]
1egoman: Fixed in 44de58c

packages/livekit-rtc/src/room.ts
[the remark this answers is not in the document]
1egoman: Fixed in 47caf82

packages/livekit-rtc/src/track.ts
```diff
@@ -21,9 +23,146 @@ export abstract class Track {
...
+      oldRoom.off('tokenRefreshed', this.onRoomTokenRefreshed);
+    }
+    if (room) {
+      room.off('tokenRefreshed', this.onRoomTokenRefreshed);
+      room.on('tokenRefreshed', this.onRoomTokenRefreshed);
+    }
+    this.roomRef = room ? new WeakRef(room) : null;
+    for (const stream of this.iterateStreams()) {
+      this.pushProcessorMetadataToStream(stream, room);
+    }
```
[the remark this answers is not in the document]
1egoman: IMO, I don't think this is a good idea. Converting these thrown errors to logs is probably unexpected by users and would mean that a failure could be easily missed. I think this is good as is.

packages/livekit-rtc/src/audio_stream_room_lifecycle.test.ts
```diff
@@ -0,0 +1,762 @@
...
+  const client = FfiClient.instance as any;
+  const origRequest = client.request;
+  const origWaitFor = client.waitFor;
+  client.request = () => ({ asyncId: BigInt(1) });
+  client.waitFor = async () => waitForResult;
+  return () => {
+    client.request = origRequest;
+    client.waitFor = origWaitFor;
+  };
+}
```
1egoman: Note to reviewers: I'm not convinced this whole testing scheme is a good idea...

The reason I ended up doing this - I want to create mock ffi class objects in memory I can use to test all this frame processor lifecycle stuff. Unfortunately, it's not really possible to disentangle into some pure functions or less coupled pieces so this to me was the least bad option I could think of.

Other thing I tried - doing some mocking with vite, but unfortunately, these tests are run on both vitest and bun and bun doesn't have support for as much of the fancy mocking features.

Maybe there's a better way to do this where there's support in the sdk itself for an "inert" ffi handle rather than working around it in the tests? Or maybe there's some better patterns I've missed that could be a better alternative?

packages/livekit-rtc/src/track.ts
```diff
@@ -21,9 +23,161 @@ export abstract class Track {
...
+  // can be constructed through Object.create in tests — bypassing the FFI handle
+  // — while keeping a stable listener identity.
+  private boundOnRoomTokenRefreshed?: () => void;
+
+  private get roomTokenRefreshedListener(): () => void {
+    if (!this.boundOnRoomTokenRefreshed) {
+      this.boundOnRoomTokenRefreshed = () => this.onRoomTokenRefreshed();
+    }
+    return this.boundOnRoomTokenRefreshed;
+  }
```
1egoman: This is also more fallout from the [above ffi client mocking stuff](https://github.com/livekit/node-sdks/pull/671/changes#r3475436398), I really am not a fan... I haven't been able to find a better solution though.

packages/livekit-rtc/src/track.ts
```diff
@@ -21,9 +23,161 @@ export abstract class Track {
   /** @internal */
   ffi_handle: FfiHandle;
 
+  private roomRef: WeakRef<Room> | null = null;
+  private audioStreams: Set<WeakRef<AudioStreamSource>> = new Set();
```
1egoman, line 27: Interesting thing I learned while building this - `WeakSet<T>` is not iterable by design.  So in this case `Set<WeakRef<T>>` is actually required given the class needs to iterate through the set to propagate events to all downstream `AudioStreamSource`s.

packages/livekit-rtc/src/track.ts
```diff
@@ -21,9 +23,161 @@ export abstract class Track {
...
+    }
+
+    if (!room) {
+      // Guard with optional-call: plugins built against an older @livekit/rtc-node
+      // inherit a FrameProcessor base class that doesn't define these methods,
+      // so they could be undefined on the prototype chain.
+      processor.onStreamInfoCleared?.();
+      processor.onCredentialsCleared?.();
+      return;
+    }
```
1egoman, line 130: Is this a worthy thing to guard against?

I'm not sure I'm aware of a pattern currently where the symbol being stored in a class with a `isXxx` type assertion pattern is being used where there's been evolution of the methods available.

theomonnom (approved):

packages/livekit-rtc/tsconfig.json
lukasIO: fine with me!

packages/livekit-rtc/src/audio_stream.ts
```diff
@@ -24,26 +31,30 @@ export interface NoiseCancellationOptions {
   options: Record<string, any>;
 }
 
-class AudioStreamSource implements UnderlyingSource<AudioFrame> {
+export class AudioStreamSource implements UnderlyingSource<AudioFrame> {
```
lukasIO, line 35: if this is only exported for testing, then let's mark it as `/** @internal */`

packages/livekit-rtc/src/track.ts
lukasIO: thanks for the explanation on why this is necessary

packages/livekit-rtc/src/track.ts
lukasIO: you mean the optional call? yeah, it doesn't hurt

lukasIO (approved):

packages/livekit-rtc/src/audio_stream.ts
1egoman: Fixed in 83203b1

1egoman:
Did some testing of this locally and in a few different happy path scenarios this all seemed to work, that plus the non happy path extensive test coverage makes me fairly confident in this change.

PR #675, opened 2026-06-10
Add `deployment` to `CreateDispatchOptions`

lukasIO:


packages/livekit-server-sdk/src/AgentDispatchClient.ts
```diff
@@ -14,15 +14,35 @@ import { ServiceBase } from './ServiceBase.js';
...
+  /** optional deployment to dispatch to. Leave empty to target the production deployment.
+   * Deployment must start and end with an alphanumeric character and may contain -, _, and . in between.
+   */
+  deployment?: string;
 }
 
 const svc = 'AgentDispatchService';
 
+/** @throws TypeError on invalid deployment names */
+export function validateAgentDeploymentString(deployment: string): void {
```
lukasIO: tbd if we even need client side validation. If we don't I'll simply remove this helper and the related tests

1egoman: If you do decide this is important (you may have more context than me on this), definitely make sure it ends up in https://github.com/livekit/client-sdk-js/pull/1971 too.

I also agree though that I think it probably makes sense to start without this, and if validation is really needed IMO it might be better to do it in upstream systems (ie, something agents hosting related?) and have it surface the validation error over webrtc.

xianshijing-lk: @1egoman , could you please remove this validateAgentDeploymentString and land the PR ?

xianshijing-lk (approved):
Please remove the validateAgentDeploymentString and relevant tests then land the PR.

packages/livekit-server-sdk/src/AgentDispatchClient.ts
1egoman: Done in 502ed86

1egoman (approved):

PR #695, opened 2026-07-04
LiveKitAPI entrypoint, ability to use a token instead of api key/secret

davidzhao:
- added a LiveKitAPI class to reduce overhead of accessing services
- ability to use token based auth instead of requiring API key/secret
- full api smoke tests to ensure correct token permissions, request encoding
- added SIPCallError
- renamed error class to ServerError


packages/livekit-server-sdk/src/LiveKitAPI.ts
```diff
@@ -0,0 +1,102 @@
...
+ * Options for {@link LiveKitAPI}. Provide either an `apiKey` and `secret`, or a
+ * pre-signed `token`. Any omitted value falls back to its environment variable.
+ */
+export interface LiveKitAPIOptions extends ClientOptions {
+  /** API key. Falls back to the `LIVEKIT_API_KEY` env var. */
+  apiKey?: string;
+  /** API secret. Falls back to the `LIVEKIT_API_SECRET` env var. */
+  secret?: string;
+  /** Pre-signed token, sent verbatim. Falls back to the `LIVEKIT_TOKEN` env var. */
+  token?: string;
```
lukasIO: is there a scenario in which we'd want to allow all three to be set together? 
If not, it would be nice if the types reflect a "either keys or token" approach

packages/livekit-server-sdk/src/LiveKitAPI.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  readonly #egress: EgressClient;
+
+  readonly #ingress: IngressClient;
+
+  readonly #sip: SipClient;
+
+  readonly #agentDispatch: AgentDispatchClient;
+
+  readonly #connector: ConnectorClient;
```
lukasIO: nit: the livekit-server-sdk doesn't use `#` as private fields. They also have a couple of downsides. Would prefer these to be regular `private` fields

packages/livekit-server-sdk/src/LiveKitAPI.ts
```diff
@@ -0,0 +1,102 @@
...
+  token?: string;
+}
+
+/**
+ * A single entry point to every LiveKit server API, exposing each service
+ * through a property, e.g. `api.room.createRoom(...)`.
+ *
+ * @example
+ * ```ts
+ * const api = new LiveKitAPI('https://my.livekit.cloud', { apiKey, secret });
```
lukasIO: nit: if all of `host`, `key` and `secret` can be set via env vars, I wonder why the API here pulls out the host to be its own parameter? (those three always come together in a bundle in our docs/sdks)

packages/livekit-server-sdk/src/ServiceBase.ts
```diff
@@ -12,20 +26,37 @@ export class ServiceBase {
...
+    this.secret = options.secret;
+    this.ttl = options.ttl || '10m';
+    this.token = options.token;
   }
 
   async authHeader(grant: VideoGrant, sip?: SIPGrant): Promise<Record<string, string>> {
+    // A pre-signed token is sent verbatim; the caller is responsible for its grants.
+    if (this.token) {
+      return { Authorization: `Bearer ${this.token}` };
+    }
```
1egoman, line 59: suggestion: It might be nice to expose here some sort of event the caller can use to determine if the token is expired or not. And if it is expired, then give the caller the ability to potentially regenerate it before actually making the API request.

This sets the stage well for a future `TokenSource` type integration (once that is officially added to node) nicely.

packages/livekit-server-sdk/src/LiveKitAPI.ts
davidzhao: agreed.. i'll set host as an option too

packages/livekit-server-sdk/src/ServiceBase.ts
davidzhao: see reply the Rust SDK PR.

I would like the server to the be the authority on token validity, instead of the client. making validity decisions on the client relies on having a correct client clock. I don't think that's a good idea.

davidzhao:
@lukasIO @1egoman feedback addressed, PTAL

packages/livekit-server-sdk/src/LiveKitAPI.ts
```diff
@@ -0,0 +1,126 @@
...
+  secret?: never;
+}
+
+/**
+ * Options for {@link LiveKitAPI}. Provide either an `apiKey` and `secret` or a
+ * pre-signed `token` — the two modes are mutually exclusive. Any omitted value
+ * falls back to its environment variable (`LIVEKIT_URL`, `LIVEKIT_API_KEY`,
+ * `LIVEKIT_API_SECRET`, `LIVEKIT_TOKEN`).
+ */
+export type LiveKitAPIOptions = LiveKitAPICommonOptions & (ApiKeyAuth | TokenAuth);
```
lukasIO, line 48: nice!

lukasIO (approved):

PR #698, opened 2026-07-23
Bump livekit-ffi to 0.12.73

1egoman:
Updates `livekit-ffi` to the latest version, which contains data streams v2.

In addition to the ffi package update itself, there's a bit more that has to be done here because the node sdk today has its own data streams implementation. I'm working on [migrating this package to use the `livekit-ffi` provided data streams](https://github.com/livekit/node-sdks/pull/697) but in the meantime, if the `livekit-ffi` upgrade were done today, then it would mean that the newly introduced `client_protocol` of `2` would be advertised, and the pre-existing node implementation would receive data stream v2 messages, not v1 (ie, single packet data streams, compression, etc could be sent to the existing node data streams code).

To address this, I've added a temporary `useLegacyClientImplementation` flag which can be set to disable advertising the new client protocol. What this means in practice:
- Outgoing data streams will all be v1 (uses node data streams implementation)
- Incoming data streams will all be v1 (other participants see `client_protocol` of `1`, so no inline content / compression)
- Outgoing RPCs will be potentially v1 or v2 (this within the rust code is powered by `livekit-ffi`'s data streams implementation, just like it works today)
- Incoming RPCs will also potentially be v1 or v2 (for the same reason)

Once https://github.com/livekit/node-sdks/pull/697 is done, I'll merge it and drop this flag both here and on the rust sdk end, and then everything will use data streams v2 everywhere.

I've tested this locally (all the data streams and rpc paths above) with success.

## Todo
- [x] Merge rust data streams v2 pull request: https://github.com/livekit/rust-sdks/pull/1192
- [x] Do `livekit-ffi` release
- [x] Bump version in here, then ci should pass
- [x] Do final round of testing

### DIFF
--- a/packages/livekit-rtc/src/index.ts
+++ b/packages/livekit-rtc/src/index.ts
@@ -36,7 +36,14 @@ export {
   TrackSource,
 } from '@livekit/rtc-ffi-bindings';
 export { VideoBufferType, VideoCodec, VideoRotation } from '@livekit/rtc-ffi-bindings';
-export { ConnectError, Room, RoomEvent, type RoomOptions, type RtcConfiguration } from './room.js';
+export {
+  ConnectError,
+  Room,
+  RoomEvent,
+  type RoomDataStreamOptions,
+  type RoomOptions,
+  type RtcConfiguration,
+} from './room.js';
 export { RpcError, type PerformRpcParams, type RpcInvocationData } from './rpc.js';
 export {
   LocalAudioTrack,
--- a/packages/livekit-rtc/src/participant.ts
+++ b/packages/livekit-rtc/src/participant.ts
@@ -277,6 +277,7 @@ export class LocalParticipant extends Participant {
     destinationIdentities?: Array<string>;
     streamId?: string;
     senderIdentity?: string;
+    totalSize?: number;
   }): Promise<TextStreamWriter> {
     const senderIdentity = options?.senderIdentity ?? this.identity;
     const streamId = options?.streamId ?? crypto.randomUUID();
@@ -287,6 +288,7 @@ export class LocalParticipant extends Participant {
       mimeType: 'text/plain',
       topic: options?.topic ?? '',
       timestamp: Date.now(),
+      totalSize: options?.totalSize,
     };
 
     const headerReq = new SendStreamHeaderRequest({
@@ -298,6 +300,7 @@ export class LocalParticipant extends Participant {
         mimeType: info.mimeType,
         topic: info.topic,
         timestamp: numberToBigInt(info.timestamp),
+        totalLength: numberToBigInt(info.totalSize),
         attributes: options?.attributes,
         contentHeader: {
           case: 'textHeader',
@@ -384,7 +387,10 @@ export class LocalParticipant extends Participant {
       streamId?: string;
     },
   ) {
-    const writer = await this.streamText(options);
+    const writer = await this.streamText({
+      ...options,
+      totalSize: new TextEncoder().encode(text).byteLength,
+    });
     await writer.write(text);
     await writer.close();
     return writer.info;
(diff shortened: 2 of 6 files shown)


1egoman:
Did a final round of light testing with a data streams v2 capable client:
- Data streams sent manually from an agent running this build are v1
- RPC requests sent manually from an agent are v2

I did some ac hoc conversation back and forth and the whole agent lifecycle seemed to work fine. I could use some help validating anything more in depth though that uses data streams, like any specific plugins / etc.

packages/livekit-rtc/src/participant.ts
```diff
@@ -384,7 +387,10 @@ export class LocalParticipant extends Participant {
       streamId?: string;
     },
   ) {
-    const writer = await this.streamText(options);
+    const writer = await this.streamText({
+      ...options,
+      totalSize: new TextEncoder().encode(text).byteLength,
```
lukasIO, line 392: this should probably also be using the `numberToBigInt` helper for consistency

pnpm-lock.yaml
lukasIO, line 1: there are some uncommitted example updates in the lock file, I think we should clean those up

packages/livekit-rtc/src/index.ts
```diff
@@ -36,7 +36,14 @@ export {
...
 export { VideoBufferType, VideoCodec, VideoRotation } from '@livekit/rtc-ffi-bindings';
-export { ConnectError, Room, RoomEvent, type RoomOptions, type RtcConfiguration } from './room.js';
+export {
+  ConnectError,
+  Room,
+  RoomEvent,
+  type RoomDataStreamOptions,
+  type RoomOptions,
+  type RtcConfiguration,
+} from './room.js';
```
1egoman, line 46: Small drive by fix, the `RoomDataStreamOptions` type wasn't exported previously.

packages/livekit-rtc/src/index.ts
```diff
@@ -36,7 +36,14 @@ export {
   TrackSource,
 } from '@livekit/rtc-ffi-bindings';
 export { VideoBufferType, VideoCodec, VideoRotation } from '@livekit/rtc-ffi-bindings';
-export { ConnectError, Room, RoomEvent, type RoomOptions, type RtcConfiguration } from './room.js';
+export {
+  ConnectError,
+  Room,
+  RoomEvent,
+  type RoomDataStreamOptions,
```
lukasIO, line 43: why should those be public?

packages/livekit-rtc/src/participant.ts
```diff
@@ -277,6 +277,7 @@ export class LocalParticipant extends Participant {
     destinationIdentities?: Array<string>;
     streamId?: string;
     senderIdentity?: string;
+    totalSize?: number;
   }): Promise<TextStreamWriter> {
```
1egoman, line 281: Other small drive by fix - `totalSize` wasn't exposed and wasn't being set on `sendText`. I suppose this is _technically_ ok but I think setting it is probably slightly better.

packages/livekit-rtc/src/participant.ts
1egoman: Discussed on slack, it's using that helper in the next layer down (`streamText`)

pnpm-lock.yaml
1egoman: Fixed, thanks, I missed this.

lukasIO (approved):

packages/livekit-rtc/src/participant.ts
```diff
@@ -298,6 +300,7 @@ export class LocalParticipant extends Participant {
         mimeType: info.mimeType,
         topic: info.topic,
         timestamp: numberToBigInt(info.timestamp),
+        totalLength: numberToBigInt(info.totalSize),
```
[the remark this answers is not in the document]
1egoman: The assumption is that clients will be implemented properly and guard against this on their own - other already existing clients will send `0` here. Also, a zero length data stream isn't something that would happen in practice right now.

theomonnom (approved):

PR #699, opened 2026-07-26
fix: cross-copy instanceof for AudioFrame and the other data classes

toubatbrian:
## Symptom

In `livekit/agents-js`, agents running with Krisp noise cancellation could not be interrupted at all, and nothing was logged. A barge-in would play out as if the user had said nothing.

The chain: the Krisp plugin loads an internal package through `createRequire`, which pulls in the **CJS** build of `@livekit/rtc-node`. The frames it handed back failed `frame instanceof AudioFrame` in the agents audio pipeline, which reaches the package through `import` and therefore holds the **ESM** build's class. Frames that fail that check are treated as unrecognised sentinel values and dropped. With no audio reaching it, the adaptive interruption classifier scored every barge-in as a backchannel.

## Mechanism

This package ships two entry points that are two separate module graphs — `dist/index.js` for the `import` condition and `dist/index.cjs` for the `require` condition. `tsup` builds them independently (`bundle: false`, both formats), so a process that reaches the package both ways gets two distinct `AudioFrame` class objects with two distinct `.prototype` objects, even though only one copy is installed on disk. `instanceof` is reference identity on the prototype chain, so it returns `false` across the two, despite identical class name, identical fields and identical source. This is Node's [dual package hazard](https://nodejs.org/api/packages.html#dual-package-hazard).

Measured during the original investigation:

```
# a frame arriving at the consumer
ctorName: "AudioFrame", sampleRate: 24000, isCjsAudioFrame: true, esmCtorIdentical: false

# isolating just the load path, no room or credentials involved
rtc-node entries in require.cache BEFORE loading the CJS-reached package: 1
rtc-node entries in require.cache AFTER:                                  27
CJS AudioFrame === ESM AudioFrame: false
CJS-built frame instanceof ESM AudioFrame: false
```

## The fix

Brand the prototype of each pure data class with a symbol from the **global** symbol registry and key `instanceof` on the brand rather than on prototype identity:

```ts
brandDataClass(AudioFrame, Symbol.for('lk.rtc-node.AudioFrame'));
```

`Symbol.for` resolves through a registry shared by every copy of a module in the process, so all copies agree on the symbol. `brandDataClass` (new, internal, `src/brand.ts`) defines the brand on the prototype as non-enumerable and installs a `static [Symbol.hasInstance]`. It is the same pattern as `Symbol.for('react.element')` and `Symbol.for('nodejs.util.inspect.custom')`, and as the existing `Symbol.for('lk.frame-processor')` in this package.

Two details in the helper worth a look:

- `null` and `undefined` are handled explicitly, so `hasInstance` cannot throw.
- Statics are inherited, so a subclass would otherwise match every branded instance of its parent. When `this` is not the branded class itself, the helper falls back to `Function.prototype[Symbol.hasInstance]`, i.e. ordinary prototype-identity behaviour. `x instanceof MySubclassOfAudioFrame` therefore still means what it used to.

### Scoping: data classes only

Branded:

| Class | Why |
| --- | --- |
| `AudioFrame` | The motivating case. Its entire contents are `data`, `sampleRate`, `channels`, `samplesPerChannel`, `_userdata`. |
| `VideoFrame` | Same shape of value type — a buffer plus dimensions and a format enum. It owns no handle: `fromOwnedInfo` copies the buffer out and disposes the FFI handle immediately. It reaches this same boundary, since `FrameProcessor<VideoFrame>` implementations are loaded from external packages exactly like the Krisp one. |
| `RpcError` | Thrown by application RPC handlers and matched with `instanceof` in `LocalParticipant.handleRpcMethodInvocation`. The handler lives in user code, which is precisely where the second copy appears; when the match fails, the caller's error code is silently replaced by `APPLICATION_ERROR`. |
| `ConnectError` | Same reasoning, from the other direction: it is caught by user code around `Room.connect`. |

**Not** branded: anything that owns an FFI handle or wraps a live resource — `Room`, `AudioStream`, `AudioSource`, `AudioResampler`, `AudioFilter`, participants, tracks, publications, E2EE types, stream readers/writers. For a value type the brand tells the whole story, because the field layout *is* the contract. For a live resource it says nothing about the things that actually have to match: handle ownership and disposal, event wiring, and the internal state the other copy's methods assume. Making `instanceof` pass there would convert a loud, immediate failure into a subtler one further downstream. This is stated in `src/brand.ts` and in a unit test that asserts those classes stay unbranded, so the reasoning does not have to be rediscovered.

Worth noting for reviewers, since it cuts the other way on the value types: `FfiClient.instance` is already stashed on `globalThis`, so both copies share one FFI client. A foreign `AudioFrame` is therefore not merely `instanceof`-compatible after this change, it is genuinely usable — verified locally that a CJS-built frame's `protoInfo()` returns correctly through the ESM copy and that `combineAudioFrames` mixes foreign and local frames.

## Verification

A unit test inside a single module instance cannot prove anything here, so there is a script that loads both published entry points in one process — `dist/index.js` via `import`, `dist/index.cjs` via `createRequire` — and checks `instanceof` in both directions. It is wired into CI right after the build step (`pnpm --filter="@livekit/rtc-node" run test:dual-package`).

**Before** (`main`, same script):

```
esm entry: .../packages/livekit-rtc/dist/index.js
cjs entry: .../packages/livekit-rtc/dist/index.cjs

ok   esm.AudioFrame !== cjs.AudioFrame: true (expected true)
ok   esm.VideoFrame !== cjs.VideoFrame: true (expected true)
ok   esm.RpcError !== cjs.RpcError: true (expected true)

FAIL cjs-built AudioFrame instanceof esm.AudioFrame: false (expected true)
FAIL esm-built AudioFrame instanceof cjs.AudioFrame: false (expected true)
FAIL cjs-built VideoFrame instanceof esm.VideoFrame: false (expected true)
FAIL esm-built VideoFrame instanceof cjs.VideoFrame: false (expected true)
FAIL cjs-built RpcError instanceof esm.RpcError: false (expected true)
FAIL esm-built RpcError instanceof cjs.RpcError: false (expected true)
FAIL cjs-built ConnectError instanceof esm.ConnectError: false (expected true)
FAIL esm-built ConnectError instanceof cjs.ConnectError: false (expected true)

ok   {} instanceof esm.AudioFrame: false (expected false)
ok   null instanceof esm.AudioFrame: false (expected false)
ok   undefined instanceof esm.AudioFrame: false (expected false)
ok   unbranded lookalike instanceof esm.AudioFrame: false (expected false)
ok   esm-built VideoFrame instanceof esm.AudioFrame: false (expected false)
ok   esm-built AudioFrame instanceof esm.VideoFrame: false (expected false)

ok   esm.Room !== cjs.Room: true (expected true)
ok   cjs Room.prototype instanceof-safe brand absent: false (expected false)

FAIL: 8 check(s) failed
```

**After** (this branch):

```
esm entry: .../packages/livekit-rtc/dist/index.js
cjs entry: .../packages/livekit-rtc/dist/index.cjs

ok   esm.AudioFrame !== cjs.AudioFrame: true (expected true)
ok   esm.VideoFrame !== cjs.VideoFrame: true (expected true)
ok   esm.RpcError !== cjs.RpcError: true (expected true)

ok   cjs-built AudioFrame instanceof esm.AudioFrame: true (expected true)
ok   esm-built AudioFrame instanceof cjs.AudioFrame: true (expected true)
ok   cjs-built VideoFrame instanceof esm.VideoFrame: true (expected true)
ok   esm-built VideoFrame instanceof cjs.VideoFrame: true (expected true)
ok   cjs-built RpcError instanceof esm.RpcError: true (expected true)
ok   esm-built RpcError instanceof cjs.RpcError: true (expected true)
ok   cjs-built ConnectError instanceof esm.ConnectError: true (expected true)
ok   esm-built ConnectError instanceof cjs.ConnectError: true (expected true)

ok   {} instanceof esm.AudioFrame: false (expected false)
ok   null instanceof esm.AudioFrame: false (expected false)
ok   undefined instanceof esm.AudioFrame: false (expected false)
ok   unbranded lookalike instanceof esm.AudioFrame: false (expected false)
ok   esm-built VideoFrame instanceof esm.AudioFrame: false (expected false)
ok   esm-built AudioFrame instanceof esm.VideoFrame: false (expected false)

ok   esm.Room !== cjs.Room: true (expected true)
ok   cjs Room.prototype instanceof-safe brand absent: false (expected false)

PASS: all checks passed
```

Note the two `esm.X !== cjs.X` lines: the classes really are distinct objects in both runs, so the `true`s in the second run come from the brand and not from the two entry points having collapsed into one.

`src/brand.test.ts` adds 35 unit tests covering, for each branded class: a genuine local instance matches; an instance from a simulated second copy matches; `{}`, `null`, `undefined`, primitives, `Object.create(null)`, functions and same-shaped unbranded lookalikes do not; an object branded as a *different* class does not; and subclasses keep prototype-identity semantics. Plus assertions that the brand stays off instances and off enumerable output, and that the handle-backed classes are left unbranded.

Results, all on this branch:

- `pnpm --filter="@livekit/rtc-node" run test` — 8 files, 104 tests passing.
- `EXPECT_BUN_RUNTIME=1 bun --bun run test` (the bun CI job) — 8 files, 104 tests passing.
- `pnpm --filter="@livekit/rtc-node" run test:dual-package` — passing, output above.
- `pnpm lint` — no new problems; the 11 remaining warnings are pre-existing and in untouched files.
- `pnpm format:check` — clean.
- `pnpm build:livekit-rtc` and `tsc` over `src/**` including the test files — clean.

## Related

Two workarounds already exist at the call sites; this PR is the root fix and neither is touched here.

- livekit/agents-js#2123 — consumer re-wraps foreign frames into its local binding.
- livekit/plugin-krisp-viva-internal#13 — producer builds frames from the caller's constructor.

## What this does not fix

Only `instanceof` for the listed classes. Two copies of the package still means two copies of the module state, and there is no attempt here to deduplicate the load itself. Code that compares constructors directly (`x.constructor === AudioFrame`) is also unaffected.

### DIFF
--- a/.changeset/cross-copy-instanceof.md
+++ b/.changeset/cross-copy-instanceof.md
@@ -1,0 +1,5 @@
+---
+'@livekit/rtc-node': patch
+---
+
+Make `instanceof` work across a dual-loaded copy of the package for the plain data classes `AudioFrame`, `VideoFrame`, `RpcError` and `ConnectError`. When a process reaches this package through both `import` (`dist/index.js`) and `require` (`dist/index.cjs`) — which happens whenever a dependency uses `createRequire` — those are two separate module graphs with two separate class objects, so `frame instanceof AudioFrame` used to be `false` for a frame built by the other copy. These classes now brand their prototype with a `Symbol.for()` key from the global symbol registry and match on the brand. Classes that own an FFI handle (`Room`, `AudioStream`, `AudioSource`, participants, tracks, ...) are deliberately left on prototype identity.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,9 @@ jobs:
         run: pnpm install
       - name: Test build
         run: pnpm --filter="*livekit*" run build
+      - name: Test rtc-node dual package instanceof
+        # Needs the built dist/, since it loads both published entry points at once.
+        run: pnpm --filter="@livekit/rtc-node" run test:dual-package
       - name: Test livekit-server-sdk (Node)
         run: pnpm --filter="livekit-server-sdk" test
       - name: Test livekit-server-sdk (Browser)
--- a/packages/livekit-rtc/package.json
+++ b/packages/livekit-rtc/package.json
@@ -54,6 +54,7 @@
     "build": "pnpm prebuild && tsup --onSuccess \"tsc --declaration --emitDeclarationOnly\"",
     "lint": "eslint -f unix \"src/**/*.ts\" --ignore-pattern \"src/proto/*\"",
     "test": "vitest run src",
+    "test:dual-package": "node scripts/verify-dual-package-instanceof.mjs",
     "test:e2e": "node scripts/run-e2e.mjs"
   }
 }
--- a/packages/livekit-rtc/scripts/verify-dual-package-instanceof.mjs
+++ b/packages/livekit-rtc/scripts/verify-dual-package-instanceof.mjs
@@ -1,0 +1,124 @@
+// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+// Loads both published entry points of this package in a single process -- `dist/index.js`
+// through `import` and `dist/index.cjs` through `createRequire` -- and checks that `instanceof`
+// works across the two copies. This is the dual package hazard that unbranded classes hit; run
+// `pnpm build` first.
+//
+//   node scripts/verify-dual-package-instanceof.mjs
+import { createRequire } from 'node:module';
+import { fileURLToPath } from 'node:url';
+import * as esm from '../dist/index.js';
+
+const require = createRequire(import.meta.url);
+const cjs = require('../dist/index.cjs');
+
+let failures = 0;
+
+const check = (label, actual, expected) => {
+  const ok = actual === expected;
+  if (!ok) failures++;
+  console.log(`${ok ? 'ok  ' : 'FAIL'} ${label}: ${actual} (expected ${expected})`);
+};
+
+console.log(`esm entry: ${fileURLToPath(new URL('../dist/index.js', import.meta.url))}`);
+console.log(`cjs entry: ${require.resolve('../dist/index.cjs')}`);
+console.log('');
+
+// Sanity check: the two entry points really are two distinct copies of the same classes.
+check('esm.AudioFrame !== cjs.AudioFrame', esm.AudioFrame !== cjs.AudioFrame, true);
+check('esm.VideoFrame !== cjs.VideoFrame', esm.VideoFrame !== cjs.VideoFrame, true);
+check('esm.RpcError !== cjs.RpcError', esm.RpcError !== cjs.RpcError, true);
+console.log('');
+
+const esmAudioFrame = esm.AudioFrame.create(48000, 1, 480);
+const cjsAudioFrame = cjs.AudioFrame.create(48000, 1, 480);
+check(
+  'cjs-built AudioFrame instanceof esm.AudioFrame',
+  cjsAudioFrame instanceof esm.AudioFrame,
+  true,
+);
+check(
+  'esm-built AudioFrame instanceof cjs.AudioFrame',
+  esmAudioFrame instanceof cjs.AudioFrame,
+  true,
+);
+
+const esmVideoFrame = new esm.VideoFrame(
+  new Uint8Array(16 * 16 * 4),
+  16,
+  16,
+  esm.VideoBufferType.RGBA,
+);
+const cjsVideoFrame = new cjs.VideoFrame(
+  new Uint8Array(16 * 16 * 4),
+  16,
+  16,
+  cjs.VideoBufferType.RGBA,
+);
+check(
+  'cjs-built VideoFrame instanceof esm.VideoFrame',
+  cjsVideoFrame instanceof esm.VideoFrame,
+  true,
+);
+check(
+  'esm-built VideoFrame instanceof cjs.VideoFrame',
+  esmVideoFrame instanceof cjs.VideoFrame,
+  true,
+);
+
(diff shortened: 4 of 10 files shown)


chenghao-mou:
From Sol:

> Blocker — collapse import and require onto one module graph.
The [exports map](https://github.com/livekit/node-sdks/blob/4fc24627dea66052e1ab184aef3e0fdd8c434fd4/packages/livekit-rtc/package.json#L10-L20) creates the identity split. Node’s documented mitigation is a single CJS implementation, optionally exposed through a thin ESM wrapper—not two independently compiled graphs. [Node.js documentation](https://nodejs.org/download/release/v20.15.0/docs/api/packages.html#approach-1-use-an-es-module-wrapper)
>
> I tested changing the import target to ./dist/index.cjs. All 55 exports remained available and identical across import/require; AudioFrame and Room identities matched on Node 18, 20, 26, and Bun 1.2.22, and an ESM TypeScript consumer compiled successfully. That removes the need for brandDataClass, the per-class whitelist, and most of the 379-line diff while also fixing duplicated module state and every constructor—not just four selected classes.
>
> The integration test should consequently load @livekit/rtc-node through its public exports in both modes. The current script hardcodes private dist paths and asserts that duplicate constructors remain distinct, encoding the hazard as a permanent invariant.
>
> High — if branding remains, scope the brand to a compatible package ABI.
[brandDataClass](https://github.com/livekit/node-sdks/blob/4fc24627dea66052e1ab184aef3e0fdd8c434fd4/packages/livekit-rtc/src/brand.ts#L39-L58) uses unversioned global keys. Future incompatible rtc-node versions will therefore accept each other’s objects. This is particularly unsafe for RpcError: passing instanceof causes LocalParticipant to call toProto(), although the brand proves neither that method nor the expected fields exist.
>
> If the package-boundary solution is rejected, have the helper accept a class identifier and internally construct a versioned/ABI-versioned Symbol.for key. The helper should enforce the global-registry invariant rather than accepting any arbitrary symbol.

PR #700, opened 2026-07-27
fix: avoid dual package hazard by building single CJS version that's …

lukasIO:
…wrapped for ESM

### DIFF
--- a/packages/livekit-rtc/scripts/postbuild.mjs
+++ b/packages/livekit-rtc/scripts/postbuild.mjs
@@ -1,7 +1,6 @@
 // SPDX-FileCopyrightText: 2024 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-
 // Runs after `tsup` (which emits the single CJS build + `.d.cts` declarations).
 // Must NOT be wired into tsup's `onSuccess`: that hook fires before tsup's own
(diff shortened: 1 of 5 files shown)


packages/livekit-rtc/scripts/postbuild.mjs
chenghao-mou, line 1: Should we have a test file with something like this?

```
import assert from 'node:assert/strict';
import { createRequire } from 'node:module';

import * as esm from '../dist/index.js';

const require = createRequire(import.meta.url);
const cjs = require('../dist/index.cjs');

assert.equal(esm.Room, cjs.Room);
assert.equal(esm.AudioFrame, cjs.AudioFrame);
```

chenghao-mou (approved):

packages/livekit-rtc/scripts/postbuild.mjs
lukasIO: good idea!

1egoman (approved):
Talked about this with lukas in a 1:1, generally makes sense. However, I think it would be ideal if it could be the other way (ie, wrap the esm version and expose as commonjs) since the majority of users should be using esm these days. If that is overly burdensome / challenging, then what is here is fine.

toubatbrian (approved):

PR #701, opened 2026-07-27
close audio streams after unsubscribe, bump FFI to 0.12.68

davidzhao:
we are still pulling in a version of the ffi prior to resume/reconnect fixes. updating the pointers to pull in these fixes.

also cleaning up the stream when unsubscribed before eos, and relaxed a few conditions to make tests less flaky

### DIFF
--- a/packages/livekit-rtc/src/audio_stream.ts
+++ b/packages/livekit-rtc/src/audio_stream.ts
@@ -123,20 +123,9 @@ export class AudioStreamSource implements UnderlyingSource<AudioFrame> {
         this.controller.enqueue(frame);
         break;
       case 'eos':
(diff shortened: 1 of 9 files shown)


packages/livekit-rtc/src/audio_stream.ts
```diff
@@ -123,20 +123,9 @@ export class AudioStreamSource implements UnderlyingSource<AudioFrame> {
         this.controller.enqueue(frame);
         break;
       case 'eos':
-        FfiClient.instance.off(FfiClientEvent.FfiEvent, this.onEvent);
```
davidzhao, line 126: refactored in teardown

paulwe (approved):

PR #704, opened 2026-07-28
fix(server-sdk): remove legacy node:crypto fallbacks to support edge/workers/isolate runtimes

intrdx:
## Summary

- Use Web Crypto and drop `await import('node:crypto')` fallbacks in `livekit-server-sdk/src/crypto/uuid.ts` and `livekit-server-sdk/src/crypto/digest.ts` 
- Bump `engines.node` to `>=19` from `>=18` (global Web Crypto support)

Workers/V8/Edge runtime bundlers fail on a dead-branch `import('node:crypto')` even when Web Crypto is available at runtime.

## Background

- #121 / #122 moved the SDK to global Web Crypto for Cloudflare Workers.
- #349 reintroduced `import('node:crypto')` in digest for Node 18. This was 2 years ago and legacy now. Node 18 went EOL 16 months ago.

This restores a Web-Crypto-only module graph for those helpers so bundlers that disallow Node builtins can use the `livekit-server-sdk`.

Refs: #121 #122 #349


intrdx:
@davidzhao, since you were really excited about this back in [PR#122](https://github.com/livekit/node-sdks/pull/122), would you mind taking a look when you have a chance? We can finally do it now to support `livekit-server-sdk` cleanly in edge/workers/isolate runtimes.

davidzhao (approved):
lgtm!

PR #708, opened 2026-08-03
update e2ee defaults and protocol

lukasIO:


1egoman (approved):
Makes sense, glad to get this fixed! Could you add a small e2e test for this as well? Something like connect two rooms with e2ee, and send a data stream from one to the other would I think have caught this

PR #710, opened 2026-08-19
fix(server-sdk): require exp claim when verifying access tokens

SashaMIT:
## Problem

`TokenVerifier.verify` calls `jose.jwtVerify` with issuer and clock tolerance only. Jose does not require `exp` unless `requiredClaims` or `maxTokenAge` is set, so a hand-rolled token with a valid signature and no expiry verifies forever.

First-party `AccessToken.toJwt` always sets `exp`. This is the same gap livekit/protocol#1706 closed on the Go verifier and livekit/python-sdks#779 on the Python verifier.

## Fix

Pass `requiredClaims: ['exp']`.

Threat-model: the attacker can mint or obtain an HS256 token that omits `exp`. They do not control the verifier secret or clock. Signature verification still succeeds and the credential never ages out. That is permanent access from a forgotten expiry, not host or agent authority.

## Test plan

- `pnpm exec vitest --environment node run src/AccessToken.test.ts` (9/9)
- Revert-tested: removing `requiredClaims: ['exp']` makes `rejects a token that omits exp` fail (the token is accepted)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/require-exp-token-verify.md
+++ b/.changeset/require-exp-token-verify.md
@@ -1,0 +1,5 @@
+---
+'livekit-server-sdk': patch
+---
+
+Require the exp claim when verifying access tokens, matching livekit/protocol#1706.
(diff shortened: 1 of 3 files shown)


frostbyte73 (approved):

SashaMIT:
Thanks for merging this, @frostbyte73. Glad the Node TokenVerifier now requires exp too.

PR #711, opened 2026-08-22
Generate request-id for api call

cnderrauber:
relate to https://github.com/livekit/server-sdk-go/pull/954

### DIFF
--- a/packages/livekit-server-sdk/src/TwirpRPC.ts
+++ b/packages/livekit-server-sdk/src/TwirpRPC.ts
@@ -186,9 +186,7 @@ export class TwirpRpc {
       'User-Agent': USER_AGENT,
       ...headers,
     };
-    if (!hasHeader(requestHeaders, REQUEST_ID_HEADER)) {
-      requestHeaders[REQUEST_ID_HEADER] = await randomUUID();
-    }
+    requestHeaders[REQUEST_ID_HEADER] = await randomUUID();
 
     const origin = new URL(this.host);
     const maxAttempts = failoverAttempts(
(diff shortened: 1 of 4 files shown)


packages/livekit-server-sdk/src/TwirpRPC.ts
```diff
@@ -175,11 +181,14 @@ export class TwirpRpc {
   ): Promise<any> {
     const path = `${this.prefix}/${this.pkg}.${service}/${method}`;
     const body = JSON.stringify(data);
-    const requestHeaders = {
+    const requestHeaders: Record<string, string> = {
       'Content-Type': 'application/json;charset=UTF-8',
       'User-Agent': USER_AGENT,
       ...headers,
     };
+    if (!hasHeader(requestHeaders, REQUEST_ID_HEADER)) {
```
lukasIO: looks like you're guarding against the possibility of the request id header being set already on the request?
I'm wondering why we're not always setting the header as part of the request headers

cnderrauber: The initial idea was to allow users to provide a `request-id`, enabling them to identify the same request when performing retries in the application.

lukasIO: I see, it doesn't exclusively relate to failover attempts then anymore. As a user I could also supply the same id to different requests, which would make things potentially awkward. I guess we would treat that as a user error.

lukasIO (approved):

packages/livekit-server-sdk/src/TwirpRPC.ts
cnderrauber: I'm not sure if it is useful to the user.. Do you think it could cause more trouble than convenience?
 The `request-id` change has not been released yet, so we can change the interface before that.

cnderrauber: Perhaps we can hold off on providing that option for now and add it later if users need it or if we see a necessity. It is always better to add an interface than to remove one from production. wdyt?

lukasIO: > Do you think it could cause more trouble than convenience?

yeah, potentially. If it's not something that users have requested so far, then we can start without it. 

> It is always better to add an interface than to remove one from production. 

💯

PR #715, opened 2026-09-01
feat(rtc): surface ParticipantActive and Participant.state

tinalenguyen:
## Problem

The FFI already emits a `ParticipantActive` event — the proto message is right there in `rtc-ffi-bindings` (`room_pb.d.ts`, `participant_active = 42`) — but `Room` drops it on the floor. Two consequences:

- JS has no way to observe a remote participant transitioning past `JOINED`.
- `Participant` exposes no `state` at all. `info.state` is set once in `createRemoteParticipant` and never refreshed, so it is not a usable substitute.

This matters because **a remote participant can only receive data messages once it reaches `ParticipantState.ACTIVE`**. Code that waits on `ParticipantConnected` gets back a participant that is present in `room.remoteParticipants` but not yet reachable, and anything sent to it is silently dropped.

The Python SDK has had `participant_active` for exactly this ([`room.py`](https://github.com/livekit/python-sdks/blob/main/livekit-rtc/livekit/rtc/room.py), "Called when a remote participant becomes active and is ready to receive data messages"). This brings Node to parity.

## Changes

**`room.ts`**
- Handle the `participantActive` FFI event: flip `info.state` to `ACTIVE` and emit the new `RoomEvent.ParticipantActive`. Unknown identity warns and no-ops, matching the existing `participantDisconnected` branch.
- Set `info.state = DISCONNECTED` alongside the existing `disconnectReason` assignment, so a departed participant's state isn't left reading `ACTIVE`.
- Declare `participantActive` in `RoomCallbacks` and `ParticipantActive` in the `RoomEvent` enum.

**`participant.ts`** — add the `state` getter, defaulting to `JOINING` since `ParticipantInfo.state` is optional in the generated types.

**`index.ts`** — re-export `ParticipantState`, next to the existing `ParticipantKind` export. Without it callers can read `.state` but have no enum to compare against.

## Compatibility

Purely additive — a new event, a new getter, a new export. No existing behavior changes: `ParticipantConnected` fires exactly as before, and nothing that ignores the new event is affected.

## Tests

Five new tests in `participant_active.test.ts`. They drive the private `onFfiEvent` handler with the same `roomEvent` shape `FfiClient` delivers, so the real switch statement is exercised rather than a reimplementation:

- state carried through `participantConnected`
- the connected → active transition, asserting no event fires on connect alone
- unknown-participant no-op
- the disconnect state flip
- the `JOINING` default when the FFI omits a state

The `FfiHandle` stub is the same one `audio_stream_room_lifecycle.test.ts` uses — fabricated handle ids otherwise trigger a native drop at GC time.

`pnpm build` succeeds; 61 tests pass across 6 files; lint is clean on the changed files.

## Motivation

Follow-up in agents-js depends on this: `waitForParticipant` there resolves on `ParticipantConnected`, so `DataStreamAudioOutput` can start streaming avatar audio to a participant that cannot yet receive it. That fix can't be written until this ships.


tinalenguyen:
Pushed `52bc5d9` — closes a gap in the first commit that a reviewer flagged.

`Participant.state` reads `ParticipantInfo.state`, and that field only ever moved on **per-participant** events. But a room-level disconnect — explicit `disconnect()` or the FFI `disconnected` event — isn't reported as each participant departing, so `cleanupOnDisconnect` left the local participant and every still-listed remote participant reporting `ACTIVE`. The new getter would then confidently report a stale state, which is worse than not having had one.

That state genuinely outlives the disconnect: the participant maps are never cleared, and `Disconnected` handlers routinely capture participants.

**Change:** `cleanupOnDisconnect` now transitions the local participant and all retained remote participants to `DISCONNECTED`, placed before `updateConnectionState` and `emit(Disconnected)` so a handler inspecting `state` sees the truth rather than racing the write.

**Tests** — four new cases in `participant_state.test.ts` (renamed from `participant_active.test.ts`, since the file now covers the whole lifecycle):
- explicit `disconnect()`, asserting on references retained from before the call
- FFI-driven `disconnected`, covering a participant that was `ACTIVE` and one that only reached `JOINED`
- ordering: states are already `DISCONNECTED` inside both `ConnectionStateChanged` and `Disconnected` handlers
- a control — `reconnecting` leaves everything `ACTIVE`

I verified the first three fail against the previous commit and pass with the fix; the control passes either way.

**One incidental change:** two remote-participant stubs in `audio_stream_room_lifecycle.test.ts` gain the `info` object that real participants always carry (`createRemoteParticipant` sets it unconditionally). Cleanup now writes through `info`, so the stubs threw. I fixed the stubs rather than adding an `info` guard to `room.ts` — a runtime check there could never fire in production, and the surrounding cleanup code already assumes participant structure.

Full suite: 65 passed, 30 skipped. Build and lint clean.

PR #723, opened 2026-09-11
failover: retry the same host when no fallback origin exists, and cover the Cloud API hosts

u9g:
## Why

A request to `cloud-api.livekit.io` was lost between Cloudflare and the origin on 2026-09-11: the edge acknowledged it, but no LiveKit system ever saw it, and the client waited out its full timeout. Every surrounding call succeeded, so a single retry would have recovered it.

The SDK's failover loop could not help, for two reasons:

- The attempts policy only matches `*.livekit.cloud`, so a `cloud-api.livekit.io` request always got exactly one attempt.
- Even with failover engaged, the loop gives up as soon as no untried origin exists, which is always the case for cloud-api: it has a single origin and `/settings/regions` returns 404.

## What

Two commits, each test-first:

1. **Retry the same host when a retryable failure has no fallback origin.** When `pickNext` finds no untried origin, a retryable failure (a thrown fetch error or an HTTP 5xx) now re-sends the request to the same host instead of surfacing it. Transport errors and 5xx are treated alike: a transport error can also mean the server executed the request and only the response was lost, so distinguishing them buys no idempotency safety, and cross-region failover already retries 5xx. Every attempt carries the same `X-Livekit-Request-Id`, so the server can dedup a replay. Max attempts and backoff are unchanged; a 4xx is still terminal.
2. **Enable retries for the LiveKit Cloud API hosts.** A new `isCloudApi` check matches `cloud-api.livekit.io` and `cloud-api.<env>.livekit.io` (case-insensitively) and joins `isCloud` in the attempts policy. A Cloud API host has a single origin, so the retry loop never fetches `/settings/regions` for it and goes straight to the same-host retry; otherwise every failed attempt would also pay the discovery timeout. The existing `.livekit.cloud` check is unchanged.

Includes a patch changeset for `livekit-server-sdk`.

## Behavior change

For cloud-api calls with a request timeout at or above `MIN_FAILOVER_TIMEOUT_SECONDS` (5s, including the 10s default), a lost request or a 5xx now costs up to `FAILOVER_MAX_ATTEMPTS` (3) attempts with exponential backoff instead of failing after one. A truly dead or persistently erroring origin takes up to three times the per-attempt budget to surface. Requests with a shorter timeout keep getting a single attempt. For `.livekit.cloud` project hosts, the same-host retry only engages once every discovered region has been tried, so their behavior is unchanged in practice.

Mirrors livekit/server-sdk-go#1002.

## Testing

`pnpm exec vitest --environment node run` in `packages/livekit-server-sdk`: 8 files, 115 tests passing, including the `test/api` integration suite against a local `livekit/test-server` container (`LK_TEST_SERVER_URL`).

New tests, each confirmed failing before its implementation commit:

- `TwirpRPC.test.ts`: "without a fallback origin, a transport error retries the same host", "without a fallback origin, a 5xx retries the same host", and "a Cloud API host retries without consulting region discovery" (asserts zero `/settings/regions` fetches).
- `failover.test.ts`: `cloud-api.livekit.io`, `cloud-api.staging.livekit.io`, and `CLOUD-API.LIVEKIT.IO` get max attempts; `cloud-api.example.com` gets one.

`pnpm lint` and `prettier --check` are clean on the changed files (the 3 existing lint warnings are pre-existing and untouched).

### DIFF
--- a/packages/livekit-server-sdk/src/TwirpRPC.ts
+++ b/packages/livekit-server-sdk/src/TwirpRPC.ts
@@ -7,6 +7,7 @@ import {
   FAILOVER_BACKOFF_BASE_MS,
   failoverAttempts,
   hostKey,
+  isCloudApi,
   pickNext,
   regionOrigins,
   sleep,
@@ -196,7 +197,8 @@ export class TwirpRpc {
       timeout,
     );
     const attempted = new Set([hostKey(origin)]);
-    let regions: string[] | undefined;
+    // A Cloud API host has a single origin; region discovery is never consulted.
+    let regions: string[] | undefined = isCloudApi(origin.hostname) ? [] : undefined;
     let current = this.host;
 
     for (let attempt = 0; attempt < maxAttempts; attempt += 1) {
--- a/packages/livekit-server-sdk/src/TwirpRPC.ts
+++ b/packages/livekit-server-sdk/src/TwirpRPC.ts
@@ -231,7 +231,8 @@ export class TwirpRpc {
         if (!regions) {
           regions = await regionOrigins(origin, headers);
         }
-        next = pickNext(regions, attempted);
+        // With no fallback origin, a retryable failure is retried against the same host.
+        next = pickNext(regions, attempted) ?? current;
       }
 
       if (!retryable || next === undefined) {
--- a/.changeset/cloud-api-failover.md
+++ b/.changeset/cloud-api-failover.md
@@ -1,0 +1,5 @@
+---
+'livekit-server-sdk': patch
+---
+
+Retry LiveKit Cloud API (`cloud-api.livekit.io`) requests on transport errors instead of failing after a single attempt.
--- a/packages/livekit-server-sdk/src/TwirpRPC.test.ts
+++ b/packages/livekit-server-sdk/src/TwirpRPC.test.ts
@@ -112,3 +112,66 @@ describe('request id', () => {
     expect(new Set(ids).size).toBe(1);
   });
 });
+
+describe('failover without a fallback origin', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  // No /settings/regions, so no fallback origin ever exists.
+  const host = 'https://cloud-api.example.com';
+
+  const okResponse = () =>
+    ({ ok: true, status: 200, json: async () => ({}) }) as unknown as Response;
+
+  const errorResponse = (status: number) =>
+    ({
+      ok: false,
+      status,
+      statusText: 'Bad Gateway',
+      headers: { get: () => null },
+      text: async () => 'bad gateway',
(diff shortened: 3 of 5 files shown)


packages/livekit-server-sdk/src/TwirpRPC.ts
```diff
@@ -231,7 +233,8 @@ export class TwirpRpc {
         if (!regions) {
           regions = await regionOrigins(origin, headers);
         }
-        next = pickNext(regions, attempted);
+        // With no fallback origin, a retryable failure is retried against the same host.
+        next = pickNext(regions, attempted) ?? current;
```
xianshijing-lk, line 237: when there is no fallback origin, what will happen to a dead origin ? with this change, it will fail after 30s rather than 10, can we do something better here ? 

As minimal, probably we should update the comment

xianshijing-lk: And my concern is, saying the service is already busy and return a 5xx (indicating busy traffic or whatever), with this change, we will retry 3 times, that might congest the traffic more.

u9g: A dead origin, ie. connection refused, DNS failure, TLS reject will fail in milliseconds.

Three attempts cost only the backoff, so ~600ms total, not 30s. The 30s worst case (10 + 0.2 + 10 + 0.4 + 10 = 30.6s at the 10s default) only happens when the origin hangs until the per-attempt timeout on every attempt. That's exactly the failure mode this PR exists to fix: on 2026-09-11 the request was acknowledged at Cloudflare's edge and never reached any LiveKit system, and the client burned its full timeout with no response. The only way to recover that is to send it again, and the only signal we have is the timeout itself.

The alternative is a single shared deadline across attempts instead of a per-attempt timeout. I don't think it belongs here: requestTimeout is documented and used as per-request today, so changing it to a total budget is a semantic change to a user-facing option, and it would have to land in server-sdk-go (already merged) and python-sdks at the same time. Happy to open that separately if you want it.

u9g: The amplification factor isn't new: FAILOVER_MAX_ATTEMPTS is 3, and cross-region failover has always retried 5xx at that same cap for .livekit.cloud hosts. This PR extends the existing 3x ceiling to single-origin hosts rather than raising it.

u9g: But the real congestion risk you're pointing at is one neither SDK guards against today: the backoff has no jitter. It's failoverBackoffMs * 2 ** attempt here and backoffBase << attempt in Go, both fully deterministic, so under a shared 5xx event every client retries in lockstep at exactly 200ms and 600ms.

Should I add jitter to the backoff?

PR #724, opened 2026-09-15
feat(rtc): add RPC interceptors

davidzhao:
Port of livekit/python-sdks#806. An RpcInterceptor wraps every RPC the local participant performs (interceptOutgoing) or handles (interceptIncoming), so telemetry can trace calls in both directions without touching user code. Interceptors run in registration order, the first outermost; errors from the remote side or from the handler flow through the chain unchanged, and a call for an unregistered method reaches the chain with `next` throwing UNSUPPORTED_METHOD, so an interceptor can record that the agent never registered it.

RpcInvocationData now carries the invoked `method`, and RpcCallInfo mirrors performRpc's parameters so an interceptor can hand a modified call to `next`.

Deliberate deviation from the Python SDK: the Node SDK never enforced the caller's responseTimeout on the handler locally (the caller times out on its side), and promises cannot be cancelled, so the incoming chain is not raced against a deadline here.


packages/livekit-rtc/README.md
```diff
@@ -121,9 +121,40 @@ try {
 
 You may find it useful to adjust the `responseTimeout` parameter, which indicates the amount of time you will wait for a response. We recommend keeping this value as low as possible while still satisfying the constraints of your application.
 
+#### Intercepting RPC calls
+
+An `RpcInterceptor` wraps every RPC the local participant performs or handles, which is useful for logging, tracing, or attaching metadata to payloads. Each method receives the call and a `next` continuation; return what `next` returns. Interceptors run in the order they were added, the first being outermost, and errors from the remote side or from your handler flow through them unchanged. Implement only the direction you care about.
```
1egoman, line 126: nitpick, but probably too late to change: In other "web framework" contexts usually this is called middleware. I wonder if that maybe would have been a better name for this?

1egoman (approved):
Generally makes sense to me, cool feature!

packages/livekit-rtc/README.md
davidzhao: they are similar but I think there are slight differences. see: [https://stackoverflow.com/questions/54863655/whats-the-difference-between-interceptor-vs-middleware-vs-filter-in-nest-js](<https://stackoverflow.com/questions/54863655/whats-the-difference-between-interceptor-vs-middleware-vs-filter-in-nest-js>)<br><br>ours is higher level with less power.. IMO interceptor is right

